PR #8, opened 2024-11-11
Install all dependencies (before : if some packages are missing, don't install any)

visill:
- [ ] Document changes


leborchuk:
Bad idea, in a that case there will be too much steps in Docker build queue. It's going to be very hard to debug.

PR #32, opened 2025-01-28
Enable copy for demo cluster

leborchuk:
We support special cloud options 
```
		ycmdb.yc_allow_copy_to_program=on
		ycmdb.yc_allow_copy_from_file=on
		ycmdb.yc_allow_copy_to_file=on
```

They are disabled by default but used in installcheck test suite. Let's enable them for the demo-cluster, which is used in our tests, and on-prem installations can also use them.

Fix test "copy" and maybe some others, not sure wich of them, lets start with test "copy"

- [ ] Document changes

### DIFF
--- a/gpAux/gpdemo/Makefile
+++ b/gpAux/gpdemo/Makefile
@@ -24,6 +24,10 @@ ifeq ($(WITH_MIRRORS), )
 WITH_MIRRORS = true
 endif
 
+ifeq ($(ENABLE_COPY), )
+ENABLE_COPY = true
+endif
+
 ifeq ($(WITH_MIRRORS), true)
 WITH_STANDBY = true
 endif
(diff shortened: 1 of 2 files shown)


gpAux/gpdemo/Makefile
```diff
@@ -24,6 +24,10 @@ ifeq ($(WITH_MIRRORS), )
 WITH_MIRRORS = true
 endif
 
+ifeq ($(ENABLE_COPY), )
```
visill, line 27: Is this how it should be?

leborchuk: Yep, enabled here by default

robozmey (approved):
Fixes tests:
- float8
- copy
- and some more (at least 5)

gpAux/gpdemo/Makefile
visill: Then why we use ifeq? Maybe just ENABLE_COPY = true

reshke (approved):

PR #72, opened 2025-02-04
Fix external_table test

leborchuk:
Let's see what's left to fix after https://github.com/open-gpdb/gpdb/commit/967a8dc2bfb2872dfd8d39e23007713ccee21bd2


leborchuk:
Conflicts with https://github.com/open-gpdb/gpdb/commit/a035bb09efc24a8294c5368b69d874ff755186e1 - it's better to fix 6.25 and then cherry-pick commits to opengpdb_stable

PR #105, opened 2025-02-14
run yezzey tests on actual ref triggered workflow

diPhantxm:
- [ ] Document changes

### DIFF
--- a/docker/yezzey/Dockerfile
+++ b/docker/yezzey/Dockerfile
@@ -78,7 +78,7 @@ RUN  git submodule update --init
 RUN rm -fr gpcontrib/yezzey 
 
 # Fetch latest yezzey version
-RUN git clone https://github.com/open-gpdb/yezzey.git gpcontrib/yezzey -b ${YEZZEY_REF} && cd /home/gpadmin 
+RUN git clone https://github.com/open-gpdb/yezzey.git gpcontrib/yezzey && git fetch origin $YEZZEY_REF:test_branch && git checkout test_branch && cd /home/gpadmin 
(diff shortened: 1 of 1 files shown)


reshke (approved):

docker/yezzey/Dockerfile
```diff
@@ -78,7 +78,7 @@ RUN  git submodule update --init
 RUN rm -fr gpcontrib/yezzey 
 
 # Fetch latest yezzey version
-RUN git clone https://github.com/open-gpdb/yezzey.git gpcontrib/yezzey -b ${YEZZEY_REF} && cd /home/gpadmin 
+RUN git clone https://github.com/open-gpdb/yezzey.git gpcontrib/yezzey && cd gpcontrib/yezzey && git fetch origin $YEZZEY_REF:test_branch && git checkout test_branch && cd /home/gpadmin
```
EinKrebs, line 81: It's kina crutchy, mb we should pass yezzeyRef arg to docker build command?

EinKrebs (changes requested):

docker/yezzey/Dockerfile
diPhantxm: We actually do it here https://github.com/open-gpdb/gpdb/blob/OPENGPDB_STABLE/docker/yezzey/Dockerfile#L7
It's just because we have ENV that takes either arg or default value

diPhantxm: I was gonna pass this arg to run tests in yezzey and yproxy repos. Perhaps, we can remove env variable from here and pass default branch explicitly

EinKrebs (approved):

PR #158, opened 2025-03-13
cherry-pick zstd fix + add guc for off changes

visill:
- [ ] Document changes

### DIFF
--- a/src/backend/storage/file/buffile.c
+++ b/src/backend/storage/file/buffile.c
@@ -60,6 +60,7 @@
 #include "postgres.h"
 
 #ifdef HAVE_LIBZSTD
+#define ZSTD_STATIC_LINKING_ONLY
 #include <zstd.h>
 #endif
 
@@ -72,6 +73,7 @@
 
 #include "cdb/cdbvars.h"
 #include "storage/gp_compress.h"
+#include "utils/gp_alloc.h"
 #include "utils/faultinjector.h"
 #include "utils/memutils.h"
 #include "utils/workfile_mgr.h"
@@ -998,6 +1000,18 @@ BufFilePledgeSequential(BufFile *buffile)
 
 #define BUFFILE_ZSTD_COMPRESSION_LEVEL 1
 
+void *
+customAlloc(void *opaque, size_t size)
+{
+	return MemoryContextAlloc(TopMemoryContext, size);
+}
+
+void
+customFree(void *opaque, void *address)
+{
+	pfree(address);
+}
+
 /*
  * Temporary buffer used during compresion. It's used only within the
  * functions, so we can allocate this once and reuse it for all files.
@@ -1012,6 +1026,10 @@ BufFileStartCompression(BufFile *file)
 {
 	ResourceOwner oldowner;
 	size_t ret;
+	ZSTD_customMem customMem;
+
+	customMem.customAlloc = customAlloc;
+	customMem.customFree = customFree;
 
 	/*
 	 * When working with compressed files, we rely on libzstd's buffer,
@@ -1038,7 +1056,7 @@ BufFileStartCompression(BufFile *file)
 	CurrentResourceOwner = file->resowner;
 
 	file->zstd_context = zstd_alloc_context();
-	file->zstd_context->cctx = ZSTD_createCStream();
+	file->zstd_context->cctx = ZSTD_createCStream_advanced(customMem);
 	if (!file->zstd_context->cctx)
 		elog(ERROR, "out of memory");
 	ret = ZSTD_initCStream(file->zstd_context->cctx, BUFFILE_ZSTD_COMPRESSION_LEVEL);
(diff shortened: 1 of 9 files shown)


src/backend/storage/file/buffile.c
```diff
@@ -998,6 +1000,18 @@ BufFilePledgeSequential(BufFile *buffile)
 
 #define BUFFILE_ZSTD_COMPRESSION_LEVEL 1
 
+void *
+customAlloc(void *opaque, size_t size)
+{
+	return MemoryContextAlloc(TopMemoryContext, size);
```
Smyatkin-Maxim: I don't like that we are polluting top-memory context here, cuz we are losing ability to debug OOM dumps here. I'd propose introducing a separate top-level memory context for zstd.

src/backend/utils/misc/guc_gp.c
```diff
@@ -3294,10 +3294,22 @@ struct config_bool ConfigureNamesBool_gp[] =
 		 NULL
 		},
 		&gp_use_synchronize_seqscans_catalog_vacuum_full,
+		true,
+		NULL, NULL, NULL
+	},
+
+	{
+		{"gp_resgroup_enable_zstd_counting", PGC_USERSET, COMPAT_OPTIONS_PREVIOUS,
```
Smyatkin-Maxim: this doesn't have anything to do with resource groups. If we turn resource groups off and use resqueues for some sick reason, memory accounting still gonna work just fine. We can name it to something like `gp_enable_zstd_memory_accounting`

Smyatkin-Maxim (changes requested):

src/backend/utils/misc/guc_gp.c
visill: done

src/backend/storage/file/buffile.c
visill: todo

src/backend/storage/file/buffile.c
```diff
@@ -1012,6 +1026,11 @@ BufFileStartCompression(BufFile *file)
 {
 	ResourceOwner oldowner;
 	size_t ret;
+	ZSTD_customMem customMem;
+	zstd_memory_context = AllocSetContextCreate(TopMemoryContext,"zstd_context", ALLOCSET_DEFAULT_SIZES);
```
Smyatkin-Maxim: Here is a leak: we create a new context each time we compress a file (that's fine), but we never delete it with MemoryContextDelete (that's not fine).

Smyatkin-Maxim (changes requested):
There is a leak that needs to be fixed.
IMO The perfect solution would be to make MemoryContext a member of BufFile. Then create context when we create BufFile and drop  the context when we close BufFile.

The problem is that we need to pass this context somehow to custommem, BUT I see that there is a member named opaque in ZSTD_customMem. I didn't find any documentation on how to use it, but I believe you can do something like `customMem.opaque = (void *)file_buf->memory_context;` and use this context from customAlloc and customFree.

This way you get rid of global context (which can't be even made static), and get rid of the leak.

src/backend/storage/file/buffile.c
```diff
@@ -1027,11 +1029,13 @@ BufFileStartCompression(BufFile *file)
 	ResourceOwner oldowner;
 	size_t ret;
 	ZSTD_customMem customMem;
-	zstd_memory_context = AllocSetContextCreate(TopMemoryContext,"zstd_context", ALLOCSET_DEFAULT_SIZES);
+	#ifdef HAVE_LIBZSTD
```
Smyatkin-Maxim: always true here?

visill: fixed

src/backend/storage/file/buffile.c
visill: changed

src/backend/storage/file/buffile.c
```diff
@@ -1026,7 +1047,7 @@ BufFileStartCompression(BufFile *file)
 	}
 
 	if (compression_buffer == NULL)
-		compression_buffer = MemoryContextAlloc(TopMemoryContext, BLCKSZ);
+		compression_buffer = MemoryContextAlloc(file->pg_zstd_context, BLCKSZ);
```
Smyatkin-Maxim: You have to use TopMemoryContext in this particular allocation as compression_buffer is a global variable here:
1. if GUC is on, you will get `use after free` error when running a second spilling query in the same connection
2. if GUC is off, you will get a crash or some weird memory error cuz there is no file->pg_zstd_context

src/backend/storage/file/buffile.c
```diff
@@ -1097,6 +1124,10 @@ BufFileEndCompression(BufFile *file)
 	ZSTD_outBuffer output;
 	size_t		ret;
 	int			wrote;
+	ZSTD_customMem customMem;
+
+	customMem.customAlloc = customAlloc;
+	customMem.customFree = customFree;
```
Smyatkin-Maxim, line 1130: customMem.opaque is forgotten here, allocations will fail.

Smyatkin-Maxim (changes requested):
There are two critical issue here with memory allocations, details are in comments.

src/backend/storage/file/buffile.c
visill: fixed

src/backend/storage/file/buffile.c
visill: fixed

Smyatkin-Maxim (approved):
LGTM now, feel free to squash and merge

PR #170, opened 2025-04-02
Fix tests in new workflow

robozmey:
- [ ] Document changes

### DIFF
--- a/.github/workflows/build-gpdb.yml
+++ b/.github/workflows/build-gpdb.yml
@@ -1190,15 +1190,10 @@ jobs:
         if: always()
         run: |
             # Search for regression.diffs recursively
-            found_files=$(find . -type f -path "*/regression.diffs")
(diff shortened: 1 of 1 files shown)


leborchuk (approved):

.github/workflows/build-gpdb.yml
```diff
@@ -1188,17 +1189,13 @@ jobs:
...
-            found_files=$(find . -type f -path "*/regression.diffs")
-            if [[ "x$found_files" != "x" ]]; then
-              for found_file in $found_files
-              do
-                echo "-----------------------------------------------------"
-                echo "Found regression.diffs at: $found_file"
-                echo "-----------------------------------------------------"
-                cat "$found_file"
-              done
+            found_file=$(find . -type f -name "regression.diffs" | head -n 1)
```
vovik0134, line 1195: Maybe it's better to loop over found files instead of taking first one (especially without ordering)?

PR #213, opened 2025-06-04
Cherry-pick: Fix plans for queries to replicated table with volatile function

NJrslv:
Another bug, not mentioned in the cherry-picked commit, was also fixed by it:

INSERT INTO distributed_table SELECT nextval() FROM replicated_table can insert
more rows than expected. This occurred because the nextval() was evaluated on 
each segment, resulting in an incorrect number of rows inserted
(3 rows instead of 1 on a 3-segment cluster).

Steps to reproduce:
"set optimizer=off;
create table dist_dst (x bigint) distributed by (x);
create table repl_src (y bigint) distributed replicated;
create sequence seq increment by 1 minvalue 1 maxvalue 42 start 1 no cycle;
insert into repl_src values (1);
insert into dist_dst (select nextval('seq') from repl_src);"

The problematic plan:
 Insert on dist_dst
   ->  Seq Scan on repl_src

Volatile functions are detected at the stage of adding the Insert node,
where the correct flow is created. Pre-existing flow request from Insert node
was being lost when the flow from its child Scan (no motion needed since
the table is replicated) overrode the parent's flow.

The fix inherits parent's flow to create appropriate motion node above and restores
it to the original one after motion creation.

Resulting Correct Plan:
 Insert on dist_dst
    ->  Redistribute Motion 1:3  (slice1; segments: 1)
          Hash Key: (nextval('seq'::regclass))
          ->  Seq Scan on repl_src

### DIFF
--- a/src/backend/cdb/cdbmutate.c
+++ b/src/backend/cdb/cdbmutate.c
@@ -845,7 +845,13 @@ apply_motion_mutator(Node *node, ApplyMotionState *context)
 	if (IsA(newnode, Motion) &&flow->req_move != MOVEMENT_NONE)
 	{
 		plan = ((Motion *) newnode)->plan.lefttree;
-		flow = plan->flow;
+
+		/* We'll recreate this motion later below. But we should save motion
+		 * request to create appropriate motion above the child node.
+		 * Original flow for the child node will be restored
+		 * after motion creation. */
+		flow->flow_before_req_move = plan->flow;
+		plan->flow = flow;
 		newnode = (Node *) plan;
 	}
 
--- a/src/test/regress/expected/rpt_optimizer.out
+++ b/src/test/regress/expected/rpt_optimizer.out
@@ -942,17 +942,17 @@ select distinct id from gp_dist_random('t_replicate_dst') order by id;
 
 -- update & delete
 explain (costs off) update t_replicate_volatile set a = 1 where b > random();
-ERROR:  could not devise a plan (cdbpath.c:2089)
+ERROR:  could not devise a plan (cdbpath.c:2092)
 explain (costs off) update t_replicate_volatile set a = 1 from t_replicate_volatile x where x.a + random() = t_replicate_volatile.b;
-ERROR:  could not devise a plan (cdbpath.c:2089)
+ERROR:  could not devise a plan (cdbpath.c:2092)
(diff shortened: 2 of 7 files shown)


src/backend/cdb/cdbmutate.c
```diff
@@ -845,7 +845,13 @@ apply_motion_mutator(Node *node, ApplyMotionState *context)
...
 	{
 		plan = ((Motion *) newnode)->plan.lefttree;
-		flow = plan->flow;
+
+		/* We'll recreate this motion later below. But we should save motion
+		 * request to create appropriate motion above the child node.
+		 * Original flow for the child node will be restored
+		 * after motion creation. */
+		flow->flow_before_req_move = plan->flow;
+		plan->flow = flow;
```
NJrslv, line 854: explain (costs off, verbose) insert into t1 (select nextval('seq') from t);

gmon.out
leborchuk, line 1: Лишний файл?

gpAux/gpdemo/gmon.out
leborchuk, line 1: лишний файл?

tmp.sql
leborchuk, line 1: Лишний файл?

src/test/regress/expected/rpt_optimizer.out
```diff
@@ -899,14 +899,35 @@ explain (costs off) select a from t_replicate_volatile union all select * from n
...
+ Insert on rpt.t_replicate_dst
    ->  Broadcast Motion 1:3  (slice1; segments: 1)
-         ->  Seq Scan on t_replicate_src
+         Output: ((nextval('t_replicate_dst_id_seq'::regclass))::integer), t_replicate_src.i
+         ->  Seq Scan on rpt.t_replicate_src
+               Output: nextval('t_replicate_dst_id_seq'::regclass), t_replicate_src.i
  Optimizer: Postgres query optimizer
-(4 rows)
+ Settings: enable_bitmapscan=off, enable_seqscan=off, optimizer=off
+(7 rows)
```
reshke, line 912: why

reshke (approved):

Smyatkin-Maxim (approved):

Smyatkin-Maxim:
@NJrslv LGTM, I'd suggest to squash your 3 commits, so there are only two commits left: yours and cherry-picked. And merge.

PR #249, opened 2025-08-01
Add completion for the DISTRIBUTED clause in CREATE TABLE

stanislavlyalin:
Add tab completion for the DISTRIBUTED keyword after CREATE TABLE with a column
list in parentheses. After DISTRIBUTED, provide completion options for
"BY(", "REPLICATED", and "RANDOMLY".

Introduce ends_with_paren() helper for detecting if a word ends with a closing
parenthesis, and use it where appropriate.

DISTRIBUTED

### DIFF
--- a/src/bin/psql/tab-complete.c
+++ b/src/bin/psql/tab-complete.c
@@ -814,6 +814,7 @@ static char *pg_strdup_keyword_case(const char *s, const char *ref);
 static PGresult *exec_query(const char *query);
 
 static void get_previous_words(int point, char **previous_words, int nwords);
+static bool ends_with_paren(const char *word);
 
 #ifdef NOT_USED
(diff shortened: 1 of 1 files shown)


src/bin/psql/tab-complete.c
```diff
@@ -4268,6 +4284,21 @@ get_previous_words(int point, char **previous_words, int nwords)
 	}
 }
 
+/*
```
reshke, line 4287: I not very exited about refactoring anything here, if not only this is absolutely in-avoidable to make sane contribution without it, because it increases diff with upstream (PostgreSQL) for no reason whatsoever. But for this case im actually fine. No action needed.

reshke:
Merged, thx!

PR #251, opened 2025-08-06
Add fts maintenance mode for cluster/DC maintenance

Smyatkin-Maxim:
This change adds a gp_fts_maintenance GUC and allows turning FTS off completely. Useful when expecting problems in DC or recovering from a DC failure.

When FTS is turned off, the failures are handled this way:

- a mirror failure will result in read-only cluster, but FTS won't change gp_segment_configuration. If a mirror[s] is started either with pg_ctl or with cluster restart, no gprecoverseg is needed - the cluster will just continue working as if nothing happened
- a primary segment failure results in cluster that is not available for both reading and writing. Again, no changed in gp_segment_configuration. Again, either restart or pg_ctl can start the failed segments and cluster will continue operating
- if there is a double fault during fts maintenance (i.e., mirror and primary with the same content are down), cluster can be fixed with either manual start of segments, or whole cluster restart. **HOWEVER** if double fault is still present during cluster start, gpstart will fail. Which is the expected begaviour. Again, no changes in gp_segment_configuration will occur.

The idea is that as soon as the infrastructure is stable, we can turn maintenance off and run gprecoverseg -Fa if necessary.

### DIFF
--- a/src/backend/utils/misc/guc_gp.c
+++ b/src/backend/utils/misc/guc_gp.c
@@ -3447,7 +3447,7 @@ struct config_bool ConfigureNamesBool_gp[] =
 
 		{
 		{"gp_fts_maintenance", PGC_SIGHUP, GP_ARRAY_TUNING,
-		 gettext_noop("Don't run FTS probes, the clustar ARRAY has known issues"),
+		 gettext_noop("Don't run FTS probes, the cluster ARRAY has known issues"),
 		 NULL
 		},
 		&gp_fts_maintenance,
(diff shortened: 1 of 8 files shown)


src/backend/utils/misc/guc_gp.c
```diff
@@ -3445,6 +3445,16 @@ struct config_bool ConfigureNamesBool_gp[] =
 		NULL, NULL, NULL
 	},
 
+		{
+		{"gp_fts_maintenance", PGC_SIGHUP, GP_ARRAY_TUNING,
+		 gettext_noop("Don't run FTS probes, the clustar ARRAY has known issues"),
```
reshke: Clustar

reshke (approved):

PR #252, opened 2025-08-11
add YAGP hooks collector query state

NJrslv:
State for constructing query key in YAGP hooks collector, 
linked with: https://github.com/open-gpdb/yagp_hooks_collector/pull/16

### DIFF
--- a/src/include/executor/execdesc.h
+++ b/src/include/executor/execdesc.h
@@ -23,14 +23,14 @@
 
 struct CdbExplain_ShowStatCtx;  /* private, in "cdb/cdbexplain.c" */
 
-typedef struct YagpHooksQueryState
+typedef struct YagpQueryKey
 {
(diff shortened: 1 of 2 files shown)


src/include/executor/execdesc.h
```diff
@@ -23,6 +23,14 @@
 
 struct CdbExplain_ShowStatCtx;  /* private, in "cdb/cdbexplain.c" */
 
+typedef struct YagpHooksQueryState
```
Smyatkin-Maxim: It's not really a state, it's a key. So the minimum change would be `YagpHooksQueryKey`, but I'd even go furhter and call it `QueryDescKey` so that it's more neutral and to the point.

Smyatkin-Maxim (changes requested):

Smyatkin-Maxim (approved):
LGTM

PR #253, opened 2025-08-19
Add submit & done query info hooks for consistency

NJrslv:
Modification places:

1. Abort in transaction
user types `BEGIN` -> user does something -> system aborts for some reason (syntax error, etc.).
`query_info_collect_hook(METRICS_QUERY_ERROR)` is not called later during abort for a previously declared cursor (e.g., `DECLARE c CURSOR FOR ...`) if an error occurs during a `FETCH`.

2. Don't call `query_info_collect_hook(METRICS_QUERY_SUBMIT)` for `skipData` queries because they don't go through executor (e.g., `CREATE TABLE ... WITH NO DATA` / `CREATE MATERIALIZED VIEW ... WITH NO DATA`).

3. `query_info_collect_hook(METRICS_QUERY_SUBMIT)` is missing for `REFRESH MATERIALIZED VIEW`.

### DIFF
--- a/src/backend/commands/createas.c
+++ b/src/backend/commands/createas.c
@@ -463,10 +463,6 @@ ExecCreateTableAs(CreateTableAsStmt *stmt, const char *queryString,
 				GetResqueueName(GetResQueueId()),
 				GetResqueuePriority(GetResQueueId()));
 	}
-
-	/* GPDB hook for collecting query info */
-	if (query_info_collect_hook)
-		(*query_info_collect_hook)(METRICS_QUERY_SUBMIT, queryDesc);
 	
 	if (into->skipData)
 	{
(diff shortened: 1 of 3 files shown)


src/backend/commands/createas.c
```diff
@@ -480,6 +476,10 @@ ExecCreateTableAs(CreateTableAsStmt *stmt, const char *queryString,
 	}
 	else
 	{
+		/* GPDB hook for collecting query info */
```
NJrslv, line 479: `skipData` queries

src/backend/commands/matview.c
```diff
@@ -478,6 +479,10 @@ refresh_matview_datafill(DestReceiver *dest, Query *query,
 				GetResqueuePriority(GetResQueueId()));
 	}
 
+	/* GPDB hook for collecting query info */
```
NJrslv, line 482: `REFRESH MATERIALIZED VIEW`

src/backend/commands/portalcmds.c
```diff
@@ -385,6 +386,10 @@ PortalCleanup(Portal portal)
 			}
 			PG_END_TRY();
 			CurrentResourceOwner = saveResourceOwner;
+		} else {
+			/* GPDB hook for collecting query info */
```
NJrslv, line 390: Abort in transaction on a utility stmt path (Executor shutdown is not called, but the submit hook is invoked. This is checked by `yagp_query_key != NULL` , which is initialized iff we went through the submit hook).

Smyatkin-Maxim (approved):

PR #264, opened 2025-09-22
Fix event triggers for partitioned tables

NJrslv:
Please see the description [here](https://github.com/postgres/postgres/commit/a2a5159ed678f0e4a0efa7e8aabbd4c03eb36843).

Repro steps:
```sql
create event trigger tr on ddl_command_end execute procedure tr_f();
create table tbl(r int) partition by range(r) (start(1) end(10) every(1));
alter table tbl add partition start(11) end(20);
```

### DIFF
--- a/src/backend/commands/event_trigger.c
+++ b/src/backend/commands/event_trigger.c
@@ -1666,11 +1666,6 @@ EventTriggerCollectSimpleCommand(ObjectAddress address,
  * Note we don't collect the command immediately; instead we keep it in
  * currentCommand, and only when we're done processing the subcommands we will
  * add it to the command list.
- *
- * XXX -- this API isn't considering the possibility of an ALTER TABLE command
- * being called reentrantly by an event trigger function.  Do we need stackable
- * commands at this level?  Perhaps at least we should detect the condition and
- * raise an error.
  */
 void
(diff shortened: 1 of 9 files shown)


src/backend/commands/event_trigger.c
```diff
@@ -1666,11 +1666,6 @@ EventTriggerCollectSimpleCommand(ObjectAddress address,
  * Note we don't collect the command immediately; instead we keep it in
  * currentCommand, and only when we're done processing the subcommands we will
  * add it to the command list.
- *
```
reshke, line 1669: why do we drop this?

reshke: ok v9.5 seems to lack this also.
https://github.com/postgres/postgres/blob/REL9_5_STABLE/src/backend/commands/event_trigger.c#L1705

reshke (approved):

src/backend/commands/event_trigger.c
NJrslv: For the sake of knowledge: 
they did stackable commands here. The problem was that we had only one global event trigger state, and previous states were not saved, resulting in clobbering on later calls. Now we have a stack of states.
=> the comment is no longer valid as they fixed it.

PR #267, opened 2025-09-29
pgaudit. Log column source on CTAS

andr-sokolov:
The pgaudit_ProcessUtility_hook function gets parsed query in the parsetree
argument. When the query is CTAS, then write to the log the name of the new
table and its columns names indicating where these columns data is taken from.

### DIFF
--- a/gpcontrib/pgaudit/pgaudit.c
+++ b/gpcontrib/pgaudit/pgaudit.c
@@ -1340,6 +1340,47 @@ pgaudit_ProcessUtility_hook(Node *parsetree,
     AuditEventStackItem *stackItem = NULL;
     int64 stackId = 0;
 
+    if (IsA(parsetree, CreateTableAsStmt))
+    {
+        CreateTableAsStmt *ctas = (CreateTableAsStmt*)parsetree;
+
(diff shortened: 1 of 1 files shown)


gpcontrib/pgaudit/pgaudit.c
```diff
@@ -1340,6 +1340,47 @@ pgaudit_ProcessUtility_hook(Node *parsetree,
     AuditEventStackItem *stackItem = NULL;
     int64 stackId = 0;
 
+    if (IsA(parsetree, CreateTableAsStmt))
+    {
+        CreateTableAsStmt *ctas = (CreateTableAsStmt*)parsetree;
+
+        ereport(auditLogClient ? auditLogLevel : COMMERROR,
```
NJrslv, line 1347: Not sure if `ereport()` alone is enough here. `log_audit_event()` does extra checks and logs more info. 
If simple logging is fine, no issues.

NJrslv: seems that we don't need such a comprehensive logging.

NJrslv (approved):

NJrslv:
Let's wait a bit for the regression tests to finish

PR #270, opened 2025-10-14
Make pgaudit tests pass

andr-sokolov:
Now pgaudit works on master only. So we don't receive messages from segments,
and the messages order is stable.
Fix use after free for auditEventStack->auditEvent.objectName.
Add DISTRIBUTED to the CREATE TABLE queries and RESOURCE QUEUE to CREATE ROLE to
get rid of Greenplum specific warnings.
Make changes in pg_hba.conf to login as users which are created in the tests.
Wrap the current user name in quotation marks, because it can contain symbols
which broke the tests, for example '-'.
Remove queries which Greenplum cannot execute: queries with more then one
writing slice and triggers updating each row.

### DIFF
--- a/.github/workflows/build-gpdb.yml
+++ b/.github/workflows/build-gpdb.yml
@@ -145,6 +145,7 @@ jobs:
               },
               {"test":"ic-gpcontrib",
                "make_configs":["gpcontrib/gp_array_agg:installcheck",
+                               "gpcontrib/gp_aux_catalog",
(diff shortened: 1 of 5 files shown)


Smyatkin-Maxim (approved):
LGTM

Smyatkin-Maxim:
One thing that stands out to me is relying on resource queues. I think it might be ok to create users w/o specifying resource queues and ignoring this output. Just in case we or someone runs this test w/ resgroups enable.

PR #272, opened 2025-10-21
pgaudit: Add logging of AST for SELECTs

andr-sokolov:
Add the ast_sel value for the pgaudit.log GUC.
When ast_sel is included in pgaudit.log and a SELECT query is executed, then
the abstract syntax tree of the query is logged.
The tree is logged before planning. The tree is useful to find out which columns
from which tables are used in SELECTs.

### DIFF
--- a/gpcontrib/pgaudit/expected/pgaudit.out
+++ b/gpcontrib/pgaudit/expected/pgaudit.out
@@ -1,6 +1,8 @@
 -- start_matchsubs
 -- m/{QUERY .*}",/
 -- s/{QUERY .*}",/{QUERY...}",/
+-- m/{QUERY .*},<none>/
+-- s/{QUERY .*},<none>/{QUERY...},<none>/
 -- m/pg_temp_\d+/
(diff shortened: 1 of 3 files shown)


NJrslv (approved):
No issues seen. Let's check the tests, then merge.
Update: Let's also confirm no memory leaks. `stack_push()` lacks `stack_pop()` and allocates memory in `CurrentMemoryContext`. If it is `TopMemoryContext` we would leak, but unlikely.

andr-sokolov:
All tests with pgaudit have passed here https://github.com/open-gpdb/gpdb/pull/273

andr-sokolov:
> Let's also confirm no memory leaks. `stack_push()` lacks `stack_pop()` and allocates memory in `CurrentMemoryContext`. If it is `TopMemoryContext` we would leak, but unlikely.

The memory is allocated in `pgaudit_planner_hook`, which is called from `planner`. The `planner` function is called in `MessageContext`, which is reset at the end of a query

PR #274, opened 2025-10-23
Add a copy of ic-good-opt-on, but with pgaudit

andr-sokolov:
When the ic-good-opt-on tests were run with pgaudit, some of them failed because
pgaudit output warnings "unrecognized node type". To make the tests comportable
with pgaudit add cases for Greenplum specific node types to GetCommandLogLevel.
GetCommandLogLevel is called only from check_log_statement which returns whether
to log a query. Queries of added types are logged as DDL.
Add the extension property to test. When this property is set, then the required
extension is created.


leborchuk (approved):
Looks good for me, now we have working test [ic-good-opt-pgaudit (pull_request)](https://github.com/open-gpdb/gpdb/actions/runs/18771765487/job/53558059283?pr=274) in github actions

PR #279, opened 2025-10-25
Don't try to use Orca for queries that it doesn't support

andr-sokolov:
Don't try to use Orca for queries that it doesn't support

On translating to DXL we can find out that the query is unsupported. The better
time to make this decision is analyse stage, because in this case we can run
Postgres planner without trying to run Orca and, as a consequence, significantly
reduce planning time.
Orca doesn't support gp_dist_random, queries on master-only tables, and function
exec locations except for any. It's easy to detect on analyse stage, that
a query contains these unsupported query properties.
The number of unsupported query properties to check before translating to DXL
will be increased later.

------------------------------
Before the patch
```
postgres=# select oid from pg_class where relname='pg_class';
 oid  
------
 1259
(1 row)

Time: 9,869 ms
postgres=# 
```

After the patch
```
postgres=# select oid from pg_class where relname='pg_class';
 oid  
------
 1259
(1 row)

Time: 1,995 ms
postgres=#
```

Don't plan simple inserts using Orca

### DIFF
--- a/src/backend/nodes/copyfuncs.c
+++ b/src/backend/nodes/copyfuncs.c
@@ -3195,6 +3195,7 @@ _copyQuery(const Query *from)
 	COPY_NODE_FIELD(constraintDeps);
 	COPY_NODE_FIELD(intoPolicy);
 	COPY_SCALAR_FIELD(parentStmtType);
+	COPY_SCALAR_FIELD(usePostgresPlanner);
 
 	return newnode;
 }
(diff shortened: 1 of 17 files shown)


leborchuk:
+1, Also we could use PG optimizer for insert values queries

Alena0704:
I think this patch looks promising. I rebased the patch and also propose to switch off Orca to prepare query plan (last commit).

PR #284, opened 2025-10-28
Write restore_command to recovery.conf

NJrslv:
#### What
Write `restore_command` to `recovery.conf` in `pg_basebackup` and `pg_rewind`. Take it from  source's `restore_command_hint`.

#### Why
`pg_basebackup`/`pg_rewind` don't write `restore_command` during `recovery.conf` generation. It can lead to timeline collisions. During promotion, a newer timeline might exist in the archive. If the promoting segment doesn't know about it, it could pick locally the same timeline ID as in the archive. Later when pushing to the archive, this causes errors that the timeline ID already exists.

#### Tests
- TAP recovery test added that reproduces collision.
- In pg_rewind tests made sure `restore_command` is written to `recovery.conf`. Without my changes `pg_xlog_symlink` cannot pass and a couple of them are flaky. *If you want to run them, make sure demo cluster is up with `export WITH_MIRRORS=true` and 
`export WITH_STANDBY=true` and do `make check-remote`


src/backend/access/transam/xlog.c
```diff
@@ -5581,6 +5581,16 @@ readRecoveryCommandFile(void)
 							item->name)));
 	}
 
+	/*
+	 * recovery.conf does not have restore_command,
+	 * fallback to GUC restore_command_hint (already loaded)
+	 */
+	if (!recoveryRestoreCommand) {
+		recoveryRestoreCommand = GetConfigOptionByName("restore_command_hint", NULL /* varname */);
```
NJrslv: Should read the guc directly (`XLogRestoreCommandDummy`)

src/backend/access/transam/xlog.c
```diff
@@ -5581,6 +5581,16 @@ readRecoveryCommandFile(void)
 							item->name)));
 	}
 
+	/*
+	 * recovery.conf does not have restore_command,
+	 * fallback to GUC restore_command_hint (already loaded)
+	 */
+	if (!recoveryRestoreCommand) {
```
NJrslv: `recoveryRestoreCommand == NULL`

src/test/recovery/t/139_restore_command_hint_timeline_collision.pl
```diff
@@ -0,0 +1,138 @@
...
+use File::Path qw(rmtree);
+
+# Initialize master with archiving and streaming
+my $node_master = get_new_node('master');
+$node_master->init(has_archiving => 1, allows_streaming => 1);
+$node_master->start;
+
+# Create test data
+$node_master->safe_psql('postgres', "CREATE TABLE test_tbl (id int)");
+$node_master->safe_psql('postgres', "INSERT INTO test_tbl VALUES (1)");
```
andr-sokolov: what about CTAS?

src/backend/access/transam/xlog.c
```diff
@@ -5581,6 +5581,16 @@ readRecoveryCommandFile(void)
 							item->name)));
 	}
 
+	/*
+	 * recovery.conf does not have restore_command,
+	 * fallback to GUC restore_command_hint (already loaded)
+	 */
+	if (recoveryRestoreCommand == NULL && XLogRestoreCommandDummy != NULL) {
```
andr-sokolov: the bracket should be on the next line according to PostgreSQL code style

andr-sokolov:
Is the added test executed on github?

NJrslv:
For history: To get recovery tests running in CI, we’d need to rewrite a bit the parser for test results in gpdb-devops  to handle both regress and prove (perl) tests. For now, I just removed the test (you can see in the artifacts it’s passing). If prove tests are needed later, we(I)’ll rewrite the parser in a separate PR.

Update: I am doing the parser thing.

src/bin/pg_rewind/sql/run_test.sh
```diff
@@ -141,20 +142,43 @@ if [ $TEST_SUITE == "local" ]; then
...
+
+    found=false
+    while IFS= read -r line; do
+        if [ "$line" = "restore_command = '/bin/true'" ]; then
+            found=true
+            break
+        fi
+    done < "$TEST_MASTER/recovery.conf"
+
+    if [ "$found" = false ]; then
```
andr-sokolov: can we use grep to simplify this code?

src/test/recovery/t/139_auto_restore_archive.pl
```diff
@@ -0,0 +1,160 @@
...
+use File::Path qw(rmtree);
+
+# Initialize master with archiving and streaming
+my $node_master = get_new_node('master');
+$node_master->init(has_archiving => 1, allows_streaming => 1);
+
+# Declare master’s archive for both master and standby
+my $archive_dir = $node_master->archive_dir;
+
+# Append restore_command_hint to master's postgres.conf
```
andr-sokolov: typo: postgresql.conf

src/test/recovery/t/139_auto_restore_archive.pl
```diff
@@ -0,0 +1,160 @@
...
+# Restore configuration
+$node_standby->append_conf('postgresql.conf', "hot_standby = off\n");
+$node_standby->append_conf('postgresql.conf', qq{
+archive_mode = on
+archive_command = '$archive_cmd'
+});
+
+# Start standby (now back on timeline 1)
+$node_standby->start;
+sleep(2);
```
andr-sokolov: what are we waiting for here? Can we replace `sleep` with check for some event? The test will be more stable

NJrslv: I think we don't need to sleep here, because start is `pg_ctl -w start`, so it waits until ready to accept connections.
Removed it.

andr-sokolov (approved):

PR #286, opened 2025-11-01
Add DSA

andr-sokolov:
DSA will be useful to dynamically allocate shared memory when the required amount is not known in advance. 

DSA will allow to get rid of GUCs in our extensions, which set the maximum shared memory amount for the extension. The extension will be easier to use, because users won't have to remember about the GUCs and set them. We will not need to handle situations where the preallocated shared memory limit is not enough.

### DIFF
--- a/src/backend/storage/ipc/dsm.c
+++ b/src/backend/storage/ipc/dsm.c
@@ -82,6 +82,7 @@ typedef struct dsm_control_item
 {
 	dsm_handle	handle;
 	uint32		refcnt;			/* 2+ = active, 1 = moribund, 0 = gone */
+	bool		pinned;
 } dsm_control_item;
 
 /* Layout of the dynamic shared memory control segment. */
@@ -490,6 +491,7 @@ dsm_create(Size size)
 			dsm_control->item[i].handle = seg->handle;
 			/* refcnt of 1 triggers destruction, so start at 2 */
 			dsm_control->item[i].refcnt = 2;
+			dsm_control->item[i].pinned = false;
 			seg->control_slot = i;
 			LWLockRelease(DynamicSharedMemoryControlLock);
 			return seg;
@@ -515,6 +517,7 @@ dsm_create(Size size)
 	dsm_control->item[nitems].handle = seg->handle;
 	/* refcnt of 1 triggers destruction, so start at 2 */
 	dsm_control->item[nitems].refcnt = 2;
+	dsm_control->item[nitems].pinned = false;
 	seg->control_slot = nitems;
 	dsm_control->nitems++;
 	LWLockRelease(DynamicSharedMemoryControlLock);
@@ -753,6 +756,9 @@ dsm_detach(dsm_segment *seg)
 		/* If new reference count is 1, try to destroy the segment. */
 		if (refcnt == 1)
 		{
+			/* A pinned segment should never reach 1. */
+			Assert(!dsm_control->item[control_slot].pinned);
+
 			/*
 			 * If we fail to destroy the segment here, or are killed before we
 			 * finish doing so, the reference count will remain at 1, which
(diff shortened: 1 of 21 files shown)


src/backend/utils/mmgr/dsa.c
```diff
@@ -0,0 +1,2214 @@
...
+		 * their caller has a double-free or access-after-free bug, which we
+		 * have no hope of detecting.  So we know it's safe to access this
+		 * array slot without holding a lock; it won't change underneath us.
+		 * Furthermore, we know that we can see the latest contents of the
+		 * slot, as explained in check_for_freed_segments, which those
+		 * functions call before arriving here.
+		 */
+		handle = area->control->segment_handles[index];
+
+		/* It's an erro to try to access an unused slot. */
```
NJrslv: erro -> error?

andr-sokolov: Fixed

src/test/modules/test_dsa/meson.build
```diff
@@ -0,0 +1,33 @@
+# Copyright (c) 2022-2023, PostgreSQL Global Development Group
```
x4m: Meson?

src/test/modules/Makefile
```diff
@@ -8,5 +8,6 @@ SUBDIRS = test_planner
 SUBDIRS += connection 
 SUBDIRS += test_extensions 
 SUBDIRS += test_bloomfilter
+SUBDIRS += test_dsa
```
x4m, line 11: Is this test included into any of our schedules?

src/backend/storage/ipc/dsm.c
```diff
@@ -822,13 +840,98 @@ dsm_pin_segment(dsm_segment *seg)
...
+{
+	uint32		control_slot = INVALID_CONTROL_SLOT;
+	bool		destroy = false;
+	uint32		i;
+
+	/* Find the control slot for the given handle. */
+	LWLockAcquire(DynamicSharedMemoryControlLock, LW_EXCLUSIVE);
+	for (i = 0; i < dsm_control->nitems; ++i)
+	{
+		/* Skip unused slots. */
```
x4m: IDK, maybe backport these?
https://github.com/postgres/postgres/commit/0b55aaacec313c309e21f63d9ff5733c3f8ab813
https://github.com/postgres/postgres/commit/6c0fb941892519ad6b8873e99c4001404fb9a128

x4m (commented):
This PR mostly brings Postgres code, so, probably this might be OK, but don't see much integration into GP stuff. Are tests actually triggered?

src/test/modules/test_dsa/meson.build
andr-sokolov: Removed

src/test/modules/Makefile
andr-sokolov: The [ic-modules](https://github.com/open-gpdb/gpdb/actions/runs/19300875539/job/55195719153?pr=286#logs) job is added

src/backend/storage/ipc/dsm.c
andr-sokolov: The https://github.com/postgres/postgres/commit/6c0fb941892519ad6b8873e99c4001404fb9a128 commit has already been backported in OPENGPDB_STABLE https://github.com/open-gpdb/gpdb/blob/94f8b86ba3a0d19e2dc1413c868d68a851be92a8/src/backend/storage/ipc/dsm.c#L574-L594.

andr-sokolov: I added 0b55aaa to the PR, because this commit fixes the same problem as https://github.com/postgres/postgres/commit/6c0fb941892519ad6b8873e99c4001404fb9a128

andr-sokolov:
The temp_tables_stat extension is implemented without DSA (https://github.com/open-gpdb/gpdb/pull/291). There is no need for DSA yet

PR #287, opened 2025-11-01
Add the temp_tables_stat extension

andr-sokolov:


gpcontrib/temp_tables_stat/temp_tables_stat.control
```diff
@@ -0,0 +1,5 @@
+# ctas_log extension
+comment = 'log columns sources'
```
NJrslv, line 2: Possible typo, shouldn't the comment be something different?

gpcontrib/temp_tables_stat/temp_tables_stat.control
```diff
@@ -0,0 +1,5 @@
...
+default_version = '0.1'
+module_pathname = '$libdir/ctas_log'
```
NJrslv, line 4: Possible typo, `$libdir/ctas_log` -> `$libdir/temp_tables_stat`

gpcontrib/temp_tables_stat/temp_tables_stat.c
```diff
@@ -0,0 +1,333 @@
...
+	funcctx = SRF_PERCALL_SETUP();
+
+	if (funcctx->call_cntr >= funcctx->max_calls)
+		SRF_RETURN_DONE(funcctx);
+
+	RelFileNode *rfns = (RelFileNode*) funcctx->user_fctx;
+	Datum		values[2] = {0};
+	bool		nulls[2] = {0};
+
+	values[0] = CStringGetTextDatum("dsadas");
```
NJrslv, line 304: We don't need the `values[0]`? I did not get the point of `dsadas`.

gpcontrib/temp_tables_stat/temp_tables_stat.c
```diff
@@ -0,0 +1,333 @@
...
+/*
+ * Cleanup temp file list.
+ * When the function is called, the list should be empty
+ */
+//static void
+//shutdown_hook(int code, Datum arg)
+//{
+//	dsa_release_in_place(backendsTmpFileArray->dsa_mem);
+//
+//	if (MyBackendId == InvalidBackendId)
```
NJrslv, line 74: Am I right that during cleanup `MyBackendId` might possibly be invalid but during execution of the other hooks it is guaranteed so? (I have this question because we currently have no bounds checking in the other hooks).

Also, why we don't need the cleanup (It is commented)?

gpcontrib/temp_tables_stat/temp_tables_stat.c
```diff
@@ -0,0 +1,333 @@
...
+	return dsa;
+}
+
+
+static void tts_file_create_hook(RelFileNodeBackend rnode)
+{
+	if (rnode.backend != TempRelBackendId)
+		return;
+
+	ereport(WARNING, (errmsg("create %s", relpath(rnode, MAIN_FORKNUM))));
```
NJrslv, line 124: Do you think `debug` would be more appropriate? (If this hook is called frequently it can flood the logs). The same applies to `tts_file_unlink_hook`, but I am not sure here.

NJrslv (changes requested):

NJrslv:
Ohh, it is a draft PR, so my comments don't make sense now.

PR #291, opened 2025-11-20
Add the temp_tables_stat extension

andr-sokolov:
This extension is useful to find out how much disk space is occupied by
temporary tables and which user created them in which session.
The extension tracks the creation and deletion of temporary table files on
segments. The list of files is stored in the shared memory of each segment.
If memory allocated using ShmemInitStruct is not enough, the extension uses DSM.
The extension adds the tts_get_seg_files function to get the tracked files,
their sizes, user and session where they was created from segments.
Add the ic-temp_tables_stat test for the extension.

### DIFF
--- a/gpcontrib/temp_tables_stat/Makefile
+++ b/gpcontrib/temp_tables_stat/Makefile
@@ -1,0 +1,20 @@
+# gpcontrib/temp_tables_stat/Makefile
+
+MODULE_big = temp_tables_stat
+OBJS = $(MODULE_big).o $(WIN32RES)
+
+EXTENSION = $(MODULE_big)
+PGFILEDESC = "Collect and show statictics on temporary tables"
+
+DATA = $(MODULE_big)--0.1.sql
+
+ifdef USE_PGXS
+PG_CONFIG = pg_config
+PGXS := $(shell $(PG_CONFIG) --pgxs)
+include $(PGXS)
+else
+subdir = gpcontrib/temp_tables_stat
+top_builddir = ../..
+include $(top_builddir)/src/Makefile.global
+include $(top_srcdir)/contrib/contrib-global.mk
+endif
--- a/gpcontrib/temp_tables_stat/Makefile
+++ b/gpcontrib/temp_tables_stat/Makefile
@@ -4,7 +4,7 @@ MODULE_big = temp_tables_stat
 OBJS = $(MODULE_big).o $(WIN32RES)
 
 EXTENSION = $(MODULE_big)
-PGFILEDESC = "Collect and show statictics on temporary tables"
+PGFILEDESC = "Collect and show statistics on temporary tables"
 
 DATA = $(MODULE_big)--0.1.sql
 
--- a/gpcontrib/temp_tables_stat/temp_tables_stat.c
+++ b/gpcontrib/temp_tables_stat/temp_tables_stat.c
@@ -288,11 +288,11 @@ tts_get_seg_files(PG_FUNCTION_ARGS)
 		oldcontext = MemoryContextSwitchTo(funcctx->multi_call_memory_ctx);
 
 		tupdesc = CreateTemplateTupleDesc(NATTR, false);
-		TupleDescInitEntry(tupdesc, (AttrNumber) 1, "user_id", OIDOID, -1, 0);
-		TupleDescInitEntry(tupdesc, (AttrNumber) 2, "sess_id", INT4OID, -1, 0);
-		TupleDescInitEntry(tupdesc, (AttrNumber) 3, "path", TEXTOID, -1, 0);
-		TupleDescInitEntry(tupdesc, (AttrNumber) 4, "content", INT2OID, -1, 0);
-		TupleDescInitEntry(tupdesc, (AttrNumber) 5, "size", INT8OID, -1, 0);
+		TupleDescInitEntry(tupdesc, 1, "user_id", OIDOID, -1, 0);
+		TupleDescInitEntry(tupdesc, 2, "sess_id", INT4OID, -1, 0);
+		TupleDescInitEntry(tupdesc, 3, "path", TEXTOID, -1, 0);
+		TupleDescInitEntry(tupdesc, 4, "content", INT2OID, -1, 0);
+		TupleDescInitEntry(tupdesc, 5, "size", INT8OID, -1, 0);
 
 		funcctx->tuple_desc = BlessTupleDesc(tupdesc);
 
--- a/gpcontrib/temp_tables_stat/temp_tables_stat.c
+++ b/gpcontrib/temp_tables_stat/temp_tables_stat.c
@@ -218,11 +218,8 @@ tts_get_file_size(const char *dirname, const char *fn_start)
 		snprintf(fn, sizeof(fn), "%s/%s", dirname, direntry->d_name);
 
 		if (stat(fn, &fst) < 0)
-		{
-			ereport(ERROR,
-					(errcode_for_file_access(),
-					 errmsg("could not stat file \"%s\": %m", fn)));
-		}
+			continue;
+
 		dirsize += fst.st_size;
 	}
 
--- a/gpcontrib/temp_tables_stat/temp_tables_stat.c
+++ b/gpcontrib/temp_tables_stat/temp_tables_stat.c
@@ -230,6 +230,37 @@ tts_get_file_size(const char *dirname, const char *fn_start)
 	return dirsize;
 }
 
+/* Copy the files info from the list to local memory */
+static RelFileNodeBackend *
+get_files(uint32 *files_num)
+{
+	RelFileNodeBackend *files;
+
+	*files_num = 0;
+
+	LWLockAcquire(&head->lock, LW_SHARED);
+
+	/* Count files of temp tables */
+	for (const TTSNode *node = &head->node; node != NULL; node = next_node(node))
+		*files_num += node->num;
+
+	/* Allocate local memory for array of the files data */
+	files = palloc(sizeof(*files) * (*files_num));
+
+	/* Combine arrays from the list nodes into one array */
+	*files_num = 0;
+	for (const TTSNode *node = &head->node; node != NULL; node = next_node(node))
+	{
+		RelFileNodeBackend *dst = files + (*files_num);
+		memcpy(dst, node->files, sizeof(*files) * node->num);
+		*files_num += node->num;
+	}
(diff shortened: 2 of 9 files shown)


gpcontrib/temp_tables_stat/Makefile
```diff
@@ -0,0 +1,20 @@
+# gpcontrib/temp_tables_stat/Makefile
+
+MODULE_big = temp_tables_stat
+OBJS = $(MODULE_big).o $(WIN32RES)
+
+EXTENSION = $(MODULE_big)
+PGFILEDESC = "Collect and show statictics on temporary tables"
```
NJrslv: statictics -> statistics

gpcontrib/temp_tables_stat/temp_tables_stat.control
```diff
@@ -0,0 +1,5 @@
+# temp_tables_stat extension
+comment = 'Collect and show statictics on temporary tables'
```
NJrslv: statictics -> statistics

gpcontrib/temp_tables_stat/temp_tables_stat.c
```diff
@@ -0,0 +1,361 @@
...
+	{
+		dsm_segment *next_seg = dsm_create(sizeof(TTSNode));
+		dsm_pin_mapping(next_seg);
+		node->next = dsm_segment_handle(next_seg);
+		node = dsm_segment_address(next_seg);
+		node->next = DSM_HANDLE_INVALID;
+		node->num = 0;
+	}
+
+	node->files[node->num++] = rnode;
```
NJrslv: Is it possible for the node to have invalid dsm handle here, in case the head of the list is initially invalid and we skip previous loop and creation of a new node, is it fine?

gpcontrib/temp_tables_stat/temp_tables_stat.c
```diff
@@ -0,0 +1,361 @@
...
+
+	if (prev_file_create_hook)
+		(*prev_file_create_hook)(rnode);
+
+	if (!RelFileNodeBackendIsTemp(rnode))
+		return;
+
+	rnode.backend = MyBackendId;
+
+	LWLockAcquire(&head->lock, LW_EXCLUSIVE);
```
NJrslv, line 114: Just to double check, `head` is allocated in `tts_shmem_startup`, is it guaranteed that `tts_file_create_hook` is called before `tts_shmem_startup` (asking in case of any edge case here)?

gpcontrib/temp_tables_stat/temp_tables_stat.c
```diff
@@ -0,0 +1,361 @@
...
+/* Get next node by current one */
+static TTSNode *
+next_node(const TTSNode *node)
+{
+	dsm_segment *dsm_seg;
+
+	if (node->next == DSM_HANDLE_INVALID)
+		return NULL;
+
+	dsm_seg = dsm_find_mapping(node->next);
```
NJrslv, line 55: Is it possible for `dsm_find_mapping` in one process to return `NULL`, and then, just before calling `dsm_attach` below, in this process, another process detaches the node, causing `dsm_attach` to return NULL and subsequently causing a `SIGSEGV` in `dsm_pin_mapping` in the first process?

gpcontrib/temp_tables_stat/temp_tables_stat.c
```diff
@@ -0,0 +1,361 @@
...
+					last_node->num--;
+				else if (last_node == &head->node)
+					head->node.num = 0;
+				else
+				{
+					/*
+					 * last_prev_node != NULL because last_node is not head.
+					 * next_node() has been called, so the mapping exists.
+					 */
+					dsm_detach(dsm_find_mapping(last_prev_node->next));
```
NJrslv: I am not so sure if it is safe to call `dsm_detach` while holding a lock, because `dsm_detach` calls user defined callbacks and so forth.

NJrslv (changes requested):

gpcontrib/temp_tables_stat/Makefile
andr-sokolov: Fixed

gpcontrib/temp_tables_stat/temp_tables_stat.control
andr-sokolov: Fixed

gpcontrib/temp_tables_stat/temp_tables_stat.c
andr-sokolov: `shmem_startup_hook` is called from `CreateSharedMemoryAndSemaphores` which creates and initializes shared memory and semaphores. This stage has been done by postmaster before backends are created. It's impossible to create or drop a table before `shmem_startup_hook`

andr-sokolov: I added the check for head to avoid the case when head is used before initialization

gpcontrib/temp_tables_stat/temp_tables_stat.c
andr-sokolov: It is impossible, because only one process modifies the list at the moment. Process should acquire the list lock before making changes

gpcontrib/temp_tables_stat/temp_tables_stat.c
andr-sokolov: user defined callbacks are not used. What are potential problems?

NJrslv (approved):

gpcontrib/temp_tables_stat/temp_tables_stat.c
```diff
@@ -0,0 +1,376 @@
...
+	LWLockAcquire(&head->lock, LW_EXCLUSIVE);
+
+	/* Don't add rnode when it exists in the list of arrays */
+	for (node = &head->node;; node = next_node(node))
+	{
+		for (int i = 0; i < node->num; i++)
+			if (RelFileNodeBackendEquals(rnode, node->files[i]))
+				goto lExit;
+
+		if (node->next == DSM_HANDLE_INVALID)
```
Smyatkin-Maxim: Why not move it to the loop condition as (node != DSM_HANDLE_INVALID)? And also I'd move this loop to a small tts_node_exists() function and remove this goto.

gpcontrib/temp_tables_stat/temp_tables_stat.c
```diff
@@ -0,0 +1,376 @@
...
+		if (strcmp(direntry->d_name, ".") == 0 ||
+			strcmp(direntry->d_name, "..") == 0 ||
+			strncmp(direntry->d_name, fn_start, fn_start_len) != 0)
+			continue;
+
+		snprintf(fn, sizeof(fn), "%s/%s", dirname, direntry->d_name);
+
+		if (stat(fn, &fst) < 0)
+		{
+			ereport(ERROR,
```
Smyatkin-Maxim: Should it really be an error? Seems quite a usual thing to happen for file to be deleted while we're scanning the directory

gpcontrib/temp_tables_stat/temp_tables_stat.c
```diff
@@ -0,0 +1,376 @@
...
+	*files_num = 0;
+
+	LWLockAcquire(&head->lock, LW_SHARED);
+
+	/* Count files of temp tables */
+	for (const TTSNode *node = &head->node; node != NULL; node = next_node(node))
+		*files_num += node->num;
+
+	/* Allocate local memory for array of the files data */
+	files = palloc(sizeof(*files) * (*files_num));
```
Smyatkin-Maxim, line 256: I wonder what happens to the lock if palloc fails with OOM handler. Not sure whether postgres leaves this lock hanging in such case or not.

gpcontrib/temp_tables_stat/temp_tables_stat.c
```diff
@@ -0,0 +1,376 @@
...
+void
+_PG_init(void)
+{
+	if (!process_shared_preload_libraries_in_progress)
+	{
+		ereport(ERROR,
+			(errmsg("temp_tables_stat is not in shared_preload_libraries")));
+	}
+
+	if (IS_QUERY_DISPATCHER())
```
Smyatkin-Maxim, line 359: An open question: what about seg-1 executor? Not sure you want to do anything on it as you only collect stats from segments anyway in your tts_get_seg_files

gpcontrib/temp_tables_stat/temp_tables_stat.c
andr-sokolov: Fixed

gpcontrib/temp_tables_stat/temp_tables_stat.c
andr-sokolov:
```c
#define IS_QUERY_DISPATCHER() (GpIdentity.segindex == MASTER_CONTENT_ID)
```
So seg-1 executor is also excluded by this condition

gpcontrib/temp_tables_stat/temp_tables_stat.c
andr-sokolov: The lock is released on any error here
```
#0  LWLockReleaseAll () at lwlock.c:1325
#1  0x000064898f1a9f52 in AbortTransaction () at xact.c:3254
#2  0x000064898f1aac96 in AbortCurrentTransaction () at xact.c:3913
#3  0x000064898f67b8f1 in PostgresMain (argc=1, argv=0x6489c8cc0aa0, dbname=0x6489c8cc0970 "postgres", username=0x6489c8cc0948 "sokolov43a-yu") at postgres.c:5114
#4  0x000064898f5d2ed5 in BackendRun (port=0x6489c8cf2fe0) at postmaster.c:4859
#5  0x000064898f5d2402 in BackendStartup (port=0x6489c8cf2fe0) at postmaster.c:4523
#6  0x000064898f5cd9c8 in ServerLoop () at postmaster.c:1953
#7  0x000064898f5cccc4 in PostmasterMain (argc=5, argv=0x6489c8cbe990) at postmaster.c:1523
#8  0x000064898f4ad0e8 in main (argc=5, argv=0x6489c8cbe990) at main.c:245
```

gpcontrib/temp_tables_stat/temp_tables_stat.c
andr-sokolov: Fixed

Smyatkin-Maxim (approved):
LGTM

PR #313, opened 2025-12-16
Add yagp_hooks_collector to gpcontrib

NJrslv:
Adds [yagp_hooks_collector](https://github.com/open-gpdb/yagp_hooks_collector) to gpcontrib.

I changed make file (gpcontrib/yagp_hooks_collector/Makefile), added a separate test gpcontrib-yagp-hooks-collector in CI that runs collector's tests.


.github/workflows/build-gpdb.yml
```diff
@@ -795,6 +801,7 @@ jobs:
 
           apt update -y
           apt install -y tzdata
+          apt install -y libprotobuf-dev protobuf-compiler
```
andr-sokolov: What about combining this line with the previous one to don't run `apt install` twice?

gpcontrib/yagp_hooks_collector/Makefile
```diff
@@ -0,0 +1,47 @@
...
+PROTO_OBJS  = $(patsubst %,src/protos/%.pb.o,$(PROTO_BASES))
+
+C_OBJS   = $(patsubst %.c,%.o,$(wildcard src/*.c src/*/*.c))
+CPP_OBJS = $(patsubst %.cpp,%.o,$(wildcard src/*.cpp src/*/*.cpp))
+OBJS     = $(C_OBJS) $(CPP_OBJS) $(PROTO_OBJS)
+
+override CFLAGS = -Wall -Wmissing-prototypes -Wpointer-arith -Wendif-labels \
+	-Wmissing-format-attribute -Wformat-security -fno-strict-aliasing -fwrapv \
+	-fexcess-precision=standard -Wno-unused-but-set-variable -Wno-address \
+	-Wno-format-truncation -Wno-stringop-truncation -g -ggdb -std=gnu99 \
```
andr-sokolov: Do we really always need debug symbols?

.github/workflows/build-gpdb.yml
NJrslv: I wanted to highlight different responsibilities: the first line is only about tzdata, and the second is about protobuf and only. Do you think I still need to merge them? :)

gpcontrib/yagp_hooks_collector/Makefile
NJrslv: I don't know if we need debug symbols in the release version.

On the one hand, since the hooks collector was drastically refactored, the symbols will help with bugs and core dumps in releases.

On the other hand, fewer compiler optimizations are performed, but I think the cost of it is negligible compared to the cost of networking accesses in Greenplum and all those things.

And most importantly there are 2-3 .c files.

What do you think?

andr-sokolov: usually CFLAGS are taken from pg_config to use the same flags as in the core. CXXFLAGS contains `g -ggdb` too

.github/workflows/build-gpdb.yml
andr-sokolov: What about one command, but two lines?
```
apt install -y tzdata \
               libprotobuf-dev protobuf-compiler
```

NJrslv: got it, thanks

gpcontrib/yagp_hooks_collector/Makefile
NJrslv: okay, what do you think if I just remove the `CFLAGS` from makefile and leave `CXXFLAGS` unchanged? Does it make sense?

andr-sokolov (approved):

NJrslv:
Please, don't merge. It's better to copy the extension as submodule and disable it by default.
Or we can leave it as it is but instead update commit's history. I am doing the latter.
Upd: now it is good to go

NJrslv:
Do we want to pollute `OPENGPDB_STABLE` branch history?

gpcontrib/yagp_hooks_collector/expected/yagp_cursors.out
```diff
@@ -55,7 +56,7 @@ RESET yagpcc.logging_mode;
 SELECT segid, query_text, query_status FROM yagpcc.log WHERE segid = -1 AND utility = true ORDER BY segid, ccnt, yagp_status_order(query_status) ASC;
  segid |                      query_text                       |    query_status     
 -------+-------------------------------------------------------+---------------------
-    -1 |                                                       | QUERY_STATUS_DONE
```
NJrslv: As you can see, the test output changed - here’s why.

`yagp_hooks_collector` sends data to another process via UDS sockets through `UDSConnector::report_query()`. If no target process exists, the function returns `false`.

When the data is sent successfully (`true` is returned from `report_query()`), large fields like the query text are cleared, otherwise, they remain. In the test, the query text now appears because, in the CI container, there’s no receiving process - so the field isn’t reset.

This behavior is intentional. We collect query data in four stages, and if one fails, later stages can still transmit large fields.

Smyatkin-Maxim (approved):

PR #321, opened 2025-12-29
gpcontrib_tests

andr-sokolov:


gpcontrib/yagp_hooks_collector/sql/yagp_cursors.sql
```diff
@@ -1,3 +1,8 @@
+-- start_ignore
+\! gpconfig -c shared_preload_libraries -v 'yagp_hooks_collector'
```
reshke, line 2: alter system

gpcontrib/Makefile
reshke, line 1: please do not refactor this, there is no value in making this unalike to 
https://github.com/apache/cloudberry/blob/main/gpcontrib/Makefile

PR #325, opened 2025-12-31
fix: gpexpand fails if GP is compiled without enable_ic_proxy

sokolval:
### This is a small PR that shall fix current gpexpand behaviour. 
Currently if you try to run **gpexpand** on a system which was compiled without --enable_icproxy  steps within gpexpand will try to query GUC gp_interconnect_proxy_addresses which would not exist on it. That is why first I moved validation that GUC **interconnect_type** is set to _proxy_.
Also put a backslash before COPY in SQL statements as it would otherwise fail with message: _ERROR:  forbidden to COPY to file in Yandex Cloud
HINT:  Anyone can COPY to stdout or from stdin. psql's \copy command also works for anyone._

### DIFF
--- a/gpMgmt/bin/gpexpand
+++ b/gpMgmt/bin/gpexpand
@@ -1845,7 +1845,7 @@ class gpexpand:
         try:
             data_file = os.path.abspath('./status_detail.dat')
             self.logger.debug('status_detail data file: %s' % data_file)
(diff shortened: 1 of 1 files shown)


reshke:
👍 👍 👍

leborchuk:
need to fix test gpexpand_gpshrink there

PR #327, opened 2026-01-09
Cache GUC values at query SUBMIT to ensure consistent filtering

NJrslv:
The extension logs `SUBMIT` and `DONE` events for each query. Previously, filtering [1] GUCs (`enable_utility`, `ignored_users_list`, etc.) were read directly at each hook (`SUBMIT`/`DONE`) call. This caused incorrect behavior when a query modified these GUCs: the query would pass filtering at `SUBMIT` but fail at `DONE` (or vice versa), leaving query state not freed.

Fix by caching GUC values once at `SUBMIT`. The cached values are used for all subsequent states (`START`, `END`, `DONE`) of that query. This ensures consistent filtering across the query lifecycle.

Implementation changes:

Convert Config from static methods to instance members
Add Config instance to EventSender, synced at SUBMIT
Pass Config reference to functions that need GUC values
Test was added in `yagp_guc_cache.sql`.

1. filtering GUC - a GUC that participates in decision on whether to process query or not, this is done in `EventSender::verify_query()`

### DIFF
--- a/gpcontrib/yagp_hooks_collector/expected/yagp_guc_cache.out
+++ b/gpcontrib/yagp_hooks_collector/expected/yagp_guc_cache.out
@@ -9,6 +9,7 @@
 -- This is due to query being actually executed between SUBMIT and DONE.
 -- start_ignore
 CREATE EXTENSION IF NOT EXISTS yagp_hooks_collector;
+SELECT yagpcc.truncate_log();
 -- end_ignore
 CREATE OR REPLACE FUNCTION print_last_query(query text)
 RETURNS TABLE(query_status text) AS $$
--- a/gpcontrib/yagp_hooks_collector/expected/yagp_guc_cache.out
+++ b/gpcontrib/yagp_hooks_collector/expected/yagp_guc_cache.out
@@ -1,0 +1,54 @@
+--
+-- Test GUC caching for query lifecycle consistency.
+--
+-- The extension logs SUBMIT and DONE events for each query.
+-- GUC values that control logging (enable_utility, ignored_users_list, ...)
+-- must be cached at SUBMIT time to ensure DONE uses the same filtering
+-- criteria. Otherwise, a SET command that modifies these GUCs would
+-- have its DONE event rejected, creating orphaned SUBMIT entries.
+-- This is due to query being actually executed between SUBMIT and DONE.
+CREATE EXTENSION yagp_hooks_collector;
+CREATE OR REPLACE FUNCTION print_last_query_like(query_pattern text)
+RETURNS TABLE(query_text text, query_status text) AS $$
+    SELECT l.query_text, l.query_status
+    FROM yagpcc.log l
+    WHERE l.segid = -1 AND l.query_text LIKE query_pattern
+    ORDER BY l.ccnt DESC
+$$ LANGUAGE sql;
+SET yagpcc.ignored_users_list TO '';
+SET yagpcc.enable TO TRUE;
+SET yagpcc.enable_utility TO TRUE;
+SET yagpcc.logging_mode TO 'TBL';
+-- SET below disables utility logging and DONE must still be logged.
+SET yagpcc.enable_utility TO FALSE;
+SELECT * FROM print_last_query_like('SET yagpcc.enable_utility%');
+             query_text              |    query_status     
+-------------------------------------+---------------------
+ SET yagpcc.enable_utility TO FALSE; | QUERY_STATUS_SUBMIT
+ SET yagpcc.enable_utility TO FALSE; | QUERY_STATUS_DONE
+(2 rows)
+
+-- SELECT below adds current user to ignore list and DONE must still be logged.
+-- start_ignore
+SELECT set_config('yagpcc.ignored_users_list', current_user, false);
+ set_config 
+------------
+ gpadmin
+(1 row)
+
+-- end_ignore
+SELECT * FROM print_last_query_like('SELECT set_config%');
+                              query_text                              |    query_status     
+----------------------------------------------------------------------+---------------------
+ SELECT set_config('yagpcc.ignored_users_list', current_user, false); | QUERY_STATUS_SUBMIT
+ SELECT set_config('yagpcc.ignored_users_list', current_user, false); | QUERY_STATUS_START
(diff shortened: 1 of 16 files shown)


gpcontrib/yagp_hooks_collector/src/EventSender.cpp
```diff
@@ -38,19 +35,19 @@ bool EventSender::verify_query(QueryDesc *query_desc, QueryState state,
...
-    // Without caching, a query that sets enable_utility to false from true
-    // would be accepted at SUBMIT (guc is true) but rejected at DONE (guc
-    // is false), causing a leak.
-    enable_utility = Config::enable_utility();
-    if (utility && enable_utility == false) {
+    // Cache GUCs once at SUBMIT. Synced GUCs are visible to all subsequent
+    // states. Without caching, a query that unsets/sets filtering GUCs would
+    // see different filter criteria at DONE, because at SUBMIT the query was
+    // not executed yet, causing DONE to be skipped/added.
+    config.sync();
```
NJrslv, line 42: The caching for later is done here, once per query.

gpcontrib/yagp_hooks_collector/expected/yagp_guc_cache.out
```diff
@@ -0,0 +1,54 @@
...
+-- have its DONE event rejected, creating orphaned SUBMIT entries.
+-- This is due to query being actually executed between SUBMIT and DONE.
+CREATE EXTENSION yagp_hooks_collector;
+CREATE OR REPLACE FUNCTION print_last_query_like(query_pattern text)
+RETURNS TABLE(query_text text, query_status text) AS $$
+    SELECT l.query_text, l.query_status
+    FROM yagpcc.log l
+    WHERE l.segid = -1 AND l.query_text LIKE query_pattern
+    ORDER BY l.ccnt DESC
+$$ LANGUAGE sql;
```
andr-sokolov:
```suggestion
CREATE OR REPLACE FUNCTION print_last_query(query text)
RETURNS TABLE(query_status text) AS $$
    SELECT query_status
    FROM yagpcc.log
    WHERE segid = -1 AND query_text = query
    ORDER BY ccnt DESC
$$ LANGUAGE sql;
```

NJrslv: one more thing, to use full name of the yagpcc.log table

gpcontrib/yagp_hooks_collector/expected/yagp_guc_cache.out
```diff
@@ -0,0 +1,54 @@
+--
+-- Test GUC caching for query lifecycle consistency.
+--
+-- The extension logs SUBMIT and DONE events for each query.
+-- GUC values that control logging (enable_utility, ignored_users_list, ...)
+-- must be cached at SUBMIT time to ensure DONE uses the same filtering
+-- criteria. Otherwise, a SET command that modifies these GUCs would
+-- have its DONE event rejected, creating orphaned SUBMIT entries.
+-- This is due to query being actually executed between SUBMIT and DONE.
+CREATE EXTENSION yagp_hooks_collector;
```
andr-sokolov: What about `create extension if not exists` and ignoring output like [here](https://github.com/open-gpdb/gpdb/blob/fb45057d1e440dd8de199cabf7e25a80eca71d06/gpcontrib/pgaudit/sql/pgaudit.sql#L18)?

gpcontrib/yagp_hooks_collector/expected/yagp_guc_cache.out
```diff
@@ -0,0 +1,54 @@
...
+CREATE OR REPLACE FUNCTION print_last_query_like(query_pattern text)
+RETURNS TABLE(query_text text, query_status text) AS $$
+    SELECT l.query_text, l.query_status
+    FROM yagpcc.log l
```
andr-sokolov: Do we need to delete all rows from yagpcc.log at the beginning and at the end of the test?

gpcontrib/yagp_hooks_collector/src/Config.cpp
```diff
@@ -16,9 +16,9 @@ static bool guc_enable_cdbstats = true;
 static bool guc_enable_collector = true;
 static bool guc_report_nested_queries = true;
 static char *guc_ignored_users = nullptr;
-static int guc_max_text_size = 1 << 20;     // in bytes (1MB)
-static int guc_max_plan_size = 1024;        // in KB
-static int guc_min_analyze_time = 10000;    // in ms
+static int guc_max_text_size = 1 << 20;  // in bytes (1MB)
+static int guc_max_plan_size = 1024;     // in KB
+static int guc_min_analyze_time = 10000; // in ms
```
andr-sokolov: This change seems to be unrelated. What about moving it to another PR?

gpcontrib/yagp_hooks_collector/src/Config.cpp
```diff
@@ -27,48 +27,17 @@ static const struct config_enum_entry logging_mode_options[] = {
...
 static void assign_ignored_users_hook(const char *, void *) {
   ignored_users_guc_dirty = true;
 }
 
-void Config::init() {
+void Config::init_gucs() {
   DefineCustomStringVariable(
       "yagpcc.uds_path", "Sets filesystem path of the agent socket", 0LL,
-      &guc_uds_path, "/tmp/yagpcc_agent.sock", PGC_SUSET,
+      &guc_uds_path, default_uds_path.data(), PGC_SUSET,
```
andr-sokolov: https://en.cppreference.com/w/cpp/string/basic_string_view/data.html
"it is typically a mistake to pass data() to a routine that takes just a const CharT* and expects a null-terminated string."

NJrslv: good catch, I can do something like this 
```c
static constexpr char default_uds_path[] = "/tmp/yagpcc_agent.sock";
```

gpcontrib/yagp_hooks_collector/expected/yagp_guc_cache.out
NJrslv: I don't think we need, the table is dropped along with extension at the end of each test. If, for some reason, the test is executing and `log` table is not empty we still won't produce incorrect results because we search for exact match - for query text.

andr-sokolov (approved):

PR #328, opened 2026-01-09
Add test for sending via UDS sockets

NJrslv:
Add regression test for `bool UDSConnector::report_query()` func which sends the data using UDS socket.


gpcontrib/yagp_hooks_collector/yagp_hooks_collector--1.1.sql
```diff
@@ -93,3 +93,18 @@ BEGIN
...
+
+CREATE FUNCTION yagpcc.__test_uds_receive(timeout_ms int DEFAULT 2000)
+RETURNS bigint
+AS 'MODULE_PATHNAME', 'yagp_test_uds_receive'
+LANGUAGE C STRICT VOLATILE EXECUTE ON MASTER;
+
+CREATE FUNCTION yagpcc.__test_uds_stop_server()
+RETURNS void
+AS 'MODULE_PATHNAME', 'yagp_test_uds_stop_server'
+LANGUAGE C STRICT VOLATILE EXECUTE ON MASTER;
```
andr-sokolov: The __test_uds_stop_server function has no arguments, so the STRICT keyword does not make sense here. I suggest to remove VOLATILE too, because functions are volatile by default

gpcontrib/yagp_hooks_collector/src/yagp_hooks_collector.c
```diff
@@ -43,3 +61,82 @@ Datum yagp_truncate_log(PG_FUNCTION_ARGS) {
...
 }
+
+Datum yagp_test_uds_start_server(PG_FUNCTION_ARGS) {
+  char *path = text_to_cstring(PG_GETARG_TEXT_PP(0));
+  struct sockaddr_un addr = {.sun_family = AF_UNIX};
+
+  if (strlen(path) >= sizeof(addr.sun_path))
+    ereport(ERROR, (errmsg("path too long")));
+
+  yagp_test_uds_stop_server(fcinfo);
```
andr-sokolov: what about calling yagp_test_uds_stop_server with NULL to be sure that will be no modifications of fcinfo?

andr-sokolov (approved):

PR #331, opened 2026-01-14
Refactor yagp_hooks_collector

NJrslv:
Refactor code:
1. Rewrite report_query(), the code is more readable and exception save, all resources are freed if exception occurs.
2. Apply clang-format.
3. Align max_{text/plan}_size types with types of their corresponding GUCs.

### DIFF
--- a/gpcontrib/yagp_hooks_collector/src/UDSConnector.cpp
+++ b/gpcontrib/yagp_hooks_collector/src/UDSConnector.cpp
@@ -28,68 +28,77 @@ static void inline log_tracing_failure(const yagpcc::SetQueryReq &req,
 bool UDSConnector::report_query(const yagpcc::SetQueryReq &req,
                                 const std::string &event,
                                 const Config &config) {
-  sockaddr_un address;
+  sockaddr_un address{};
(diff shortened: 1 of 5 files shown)


gpcontrib/yagp_hooks_collector/src/UDSConnector.cpp
```diff
@@ -28,68 +28,77 @@ static void inline log_tracing_failure(const yagpcc::SetQueryReq &req,
...
-      ereport(WARNING,
-              (errmsg("Unable to create non-blocking socket connection %s",
-                      strerror(errno))));
-      success = false;
-      YagpStat::report_error();
-    }
-    close(sockfd);
-  } else {
-    // log the error and go on
+  std::strcpy(address.sun_path, uds_path.c_str());
```
andr-sokolov: do we really need `std::` here?

NJrslv: It is not mandatory, we can use the C function instead. To show clearer intent for C++ here, I used the C++ version.

andr-sokolov (approved):

PR #348, opened 2026-01-21
Run resource groups tests

andr-sokolov:
Resource groups require cgroup v1, but github provides v2. So the tests are run
in QEMU to pass arguments to Linux kernel to use cgroup v1.
Don't link pg_isolation2_regress against redundant libraries, so as not to
install them in QEMU.
Fix the resgroup_cpuset test: plpythonu and gp_inject_fault may exist before
the test starts.
Comment the resgroup_cpu_rate_limit test, because it is broken.

### DIFF
--- a/.github/workflows/build-gpdb.yml
+++ b/.github/workflows/build-gpdb.yml
@@ -719,7 +719,7 @@ jobs:
           } >> "$GITHUB_STEP_SUMMARY" || true
 
   resgroup-test:
-    name: Resgroup test
+    name: isolation2 installcheck-resgroup
     needs: [prepare-build-matrix, build]
     runs-on: ubuntu-latest
     timeout-minutes: 120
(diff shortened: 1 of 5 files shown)


.github/workflows/build-gpdb.yml
```diff
@@ -718,6 +718,163 @@ jobs:
...
+          set -eux
+          PIDFILE=qemu.pid
+          if [ -f $PIDFILE ]; then
+            # Read PID safely
+            pid=$(cat $PIDFILE) && {
+              # Graceful shutdown first
+              sudo kill $pid 2>/dev/null || true
+              sleep 3
+              # Check if process still exists with the same PID and name
+              if ps -p $pid >/dev/null && [ "$(ps -p $pid -o comm=)" = "qemu-system-x86" ]; then
```
NJrslv, line 851: Shouldn't `qemu-system-x86` be `qemu-system-x86_64`?

andr-sokolov: Process name in the comm field is truncated
```
root@sokolov43a-yu1:/home/sokolov43a-yu/test_ci/vm-work# ps -fp $pid
UID          PID    PPID  C STIME TTY          TIME CMD
root      679962  679961  9 10:18 pts/3    00:00:19 qemu-system-x86_64 -machine type=pc,accel=kvm -cpu host -smp 4 -m 8192 -pidfile qemu.pid -drive file=vm-disk.qcow2,format=qcow2,if=virtio -drive file=seed.
root@sokolov43a-yu1:/home/sokolov43a-yu/test_ci/vm-work# ps -p $pid -o comm=
qemu-system-x86
root@sokolov43a-yu1:/home/sokolov43a-yu/test_ci/vm-work# 
```

.github/workflows/build-gpdb.yml
```diff
@@ -718,6 +718,163 @@ jobs:
...
+
+              cd /home/gpadmin/gpdb;
+              make create-demo-cluster;
+              source gpAux/gpdemo/gpdemo-env.sh;
+              gpconfig -c gp_resource_manager -v group;
+              gpstop -ar;
+
+              make -C src/test/regress;
+              make -C src/test/isolation2 installcheck-resgroup;
+              echo $? > /home/gpadmin/exit_code'
```
NJrslv: I think we should escape `$?`. Look at the example below:
```bash
❯ echo $VAR
HOST
❯ cat > b << EOF0
heredoc> 'echo $VAR'
heredoc> EOF0
❯ cat b
'echo HOST'
```

NJrslv (changes requested):

NJrslv (approved):

PR #354, opened 2026-02-05
Fix segfault when two SIGTERMs come

andr-sokolov:
When WAL receiver gets SIGTERM the WalRcvShutdownHandler function is called
where the got_SIGTERM flag is set. Then got_SIGTERM is checked in
ProcessWalRcvInterrupts and FATAL is passed to ereport. It leads to calling of
proc_exit and AuxiliaryProcKill where MyProc become NULL.

Segfault occurred in SetLatch called from WalRcvShutdownHandler when SIGTERM
came again and MyProc had already been NULL.

Don't call SetLatch when MyProc is NULL. Save errno only when SetLatch is called
because errno doesn't change otherwise.

### DIFF
--- a/src/backend/replication/walreceiver.c
+++ b/src/backend/replication/walreceiver.c
@@ -777,19 +777,26 @@ WalRcvSigUsr1Handler(SIGNAL_ARGS)
 static void
 WalRcvShutdownHandler(SIGNAL_ARGS)
 {
-	int			save_errno = errno;
-
 	got_SIGTERM = true;
 
 	/*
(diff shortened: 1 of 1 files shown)


NJrslv (approved):

leborchuk (approved):
В PG их выпилили - https://github.com/postgres/postgres/commit/1a2ae7c50fb4ab17a8d38e755fc871ebc3a5845d Но это требует слишком большого рефактринга ) If нужен, в коде PG он тоже был

PR #357, opened 2026-02-17
Fix gpexpand not changing content-id for wal-g cmd

NJrslv:
When expanding a cluster, gpexpand copies the postgresql.conf file directly from the template segment (content 0). This causes issues for tools like wal-g which use a --content-id flag in archive_command and restore_command_hint.

Previously, new segments inherited --content-id=0 from the template. This caused them to push WAL segments to the wrong location, potentially overwriting segment 0's segments.

This fix ensures the content ID in archive_command and restore_command_hint is updated to match the new segment's ID during expansion. If the commands do not contain the --content-id flag, they remain unchanged.

### DIFF
--- a/src/test/isolation2/expected/gpexpand_archive_restore_conf.out
+++ b/src/test/isolation2/expected/gpexpand_archive_restore_conf.out
@@ -1,0 +1,966 @@
+-- Verify that gpexpand correctly updates segment-specific flags for wal-g in
+-- archive_command and restore_command_hint when initializing new segments.
+--
+-- Previously, these GUCs were copied verbatim from the template segment
+-- (content 0), causing new segments to invoke archiving with an incorrect
+-- content-id (possible overwriting archive or restoring other's data).
+
+-- Cleanup any previous state
+!\retcode yes | gpexpand -c;
+-- start_ignore
+20260217:07:55:15:867274 gpexpand:da6db2bb5ee4:gpadmin-[INFO]:-local Greenplum Version: 'postgres (Greenplum Database) 6.28.8-python+dev.197.g05e6484876 build dev'
+20260217:07:55:16:867274 gpexpand:da6db2bb5ee4:gpadmin-[INFO]:-master Greenplum Version: 'PostgreSQL 9.4.26 (Greenplum Database 6.28.8-python+dev.197.g05e6484876 build dev) on x86_64-unknown-linux-gnu, compiled by gcc (Ubuntu 11.4.0-1ubuntu1~22.04.2) 11.4.0, 64-bit compiled on Feb 12 2026 14:27:36 (with assert checking)'
+20260217:07:55:16:867274 gpexpand:da6db2bb5ee4:gpadmin-[INFO]:-Querying gpexpand schema for current expansion state
+
+System Expansion is used to add segments to an existing GPDB array.
+gpexpand did not detect a System Expansion that is in progress.
+
+Before initiating a System Expansion, you need to provision and burn-in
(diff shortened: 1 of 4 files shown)


ostinru (approved):
🚀

src/test/isolation2/expected/gpexpand_archive_restore_conf.out
```diff
@@ -0,0 +1,966 @@
...
+-- archive_command and restore_command_hint when initializing new segments.
+--
+-- Previously, these GUCs were copied verbatim from the template segment
+-- (content 0), causing new segments to invoke archiving with an incorrect
+-- content-id (possible overwriting archive or restoring other's data).
+
+-- Cleanup any previous state
+!\retcode yes | gpexpand -c;
+-- start_ignore
+20260217:07:55:15:867274 gpexpand:da6db2bb5ee4:gpadmin-[INFO]:-local Greenplum Version: 'postgres (Greenplum Database) 6.28.8-python+dev.197.g05e6484876 build dev'
```
visill, line 11: shouldnt pass

src/test/isolation2/expected/gpexpand_archive_restore_conf.out
```diff
@@ -0,0 +1,966 @@
...
+20260217:07:55:19:867781 gpexpand:da6db2bb5ee4:gpadmin-[INFO]:-local Greenplum Version: 'postgres (Greenplum Database) 6.28.8-python+dev.197.g05e6484876 build dev'
+20260217:07:55:19:867781 gpexpand:da6db2bb5ee4:gpadmin-[INFO]:-master Greenplum Version: 'PostgreSQL 9.4.26 (Greenplum Database 6.28.8-python+dev.197.g05e6484876 build dev) on x86_64-unknown-linux-gnu, compiled by gcc (Ubuntu 11.4.0-1ubuntu1~22.04.2) 11.4.0, 64-bit compiled on Feb 12 2026 14:27:36 (with assert checking)'
+20260217:07:55:19:867781 gpexpand:da6db2bb5ee4:gpadmin-[INFO]:-Querying gpexpand schema for current expansion state
+20260217:07:55:20:867781 gpexpand:da6db2bb5ee4:gpadmin-[INFO]:-Successfully finished pg_controldata /home/gpadmin/gpdb/gpAux/gpdemo/datadirs/dbfast1/demoDataDir0 for dbid 2:
+stdout: pg_control version number:            9420600
+Catalog version number:               301908232
+Database system identifier:           7607734725579485657
+Database cluster state:               in production
+pg_control last modified:             Tue 17 Feb 2026 07:55:15 AM GMT
+Latest checkpoint location:           0/C2E0168
```
visill, line 97: and i think this will be diffrent too

visill (dismissed):
Test out is bad.

src/test/isolation2/expected/gpexpand_archive_restore_conf.out
NJrslv: I think it should be fine. This `\retcode` is expanded into something that wraps the next command in `-- start_ignore` and `-- end_ignore`.

src/test/isolation2/expected/gpexpand_archive_restore_conf.out
NJrslv: It looks like all's good:
```
============== running regression test queries        ==============
test gpexpand_gpshrink        ... ok (90.18 sec)  (diff:0.07 sec)
test gpexpand_archive_restore_conf ... ok (27.42 sec)  (diff:0.05 sec)

=====================
 All 2 tests passed. 
=====================
```

PR #359, opened 2026-03-03
Add oids mappings with objects names

andr-sokolov:
It is necessary to know names of tables and functions used in a query, but AST
contains oids only. We cannot use the system catalog tables to find out
the names, because tables and functions may have been dropped before the logs
are analysed.

Append oids mappings with names of relations and functions to AST. These
mappings also contain list of columns to get column name by its index.

The nodeToString function is replaced with query_to_text. This new function
finds out relations and functions oids in the Query structure. If they exist,
the function appends records about them to the output of nodeToString.
Otherwise, query_to_text return value is the same as that of nodeToString.

### DIFF
--- a/gpcontrib/pgaudit/pgaudit.c
+++ b/gpcontrib/pgaudit/pgaudit.c
@@ -23,6 +23,7 @@
 #include "commands/event_trigger.h"
 #include "executor/executor.h"
 #include "executor/spi.h"
+#include "funcapi.h"
 #include "miscadmin.h"
 #include "libpq/auth.h"
 #include "nodes/nodes.h"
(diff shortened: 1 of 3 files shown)


gpcontrib/pgaudit/pgaudit.c
```diff
@@ -1501,6 +1502,181 @@ pgaudit_object_access_hook(ObjectAccessType access,
...
+}
+
+static void
+append_rel_attrs(StringInfo str, Oid relOid, int16 natts)
+{
+    for (AttrNumber n = 1; n <= natts; n++)
+    {
+        char *att = get_attname(relOid, n);
+        if (att == NULL)
+            break;
```
NJrslv, line 1627: Shouldn't we write continue instead of break? Is it possible for i-th attribute be null but (i+1)-st being non-null?

gpcontrib/pgaudit/pgaudit.c
```diff
@@ -1501,6 +1502,181 @@ pgaudit_object_access_hook(ObjectAccessType access,
...
+
+                    if (argnames[i] != NULL)
+                        appendStringInfoString(str, argnames[i]);
+
+                    printComma = true;
+                }
+        }
+
+        if (argnames != NULL)
+            pfree(argnames);
```
NJrslv, line 1580: Isn't argnames an array? I mean don't we need to loop over argnames and pfree(argnames[i])?

gpcontrib/pgaudit/pgaudit.c
```diff
@@ -1501,6 +1502,181 @@ pgaudit_object_access_hook(ObjectAccessType access,
...
+
+    if (IsA(node, RangeTblEntry))
+    {
+        RangeTblEntry *rte = (RangeTblEntry *) node;
+        if (rte->rtekind == RTE_RELATION)
+            *oids = list_append_unique_oid(*oids, rte->relid);
+        return false;
+    }
+
+    if (IsA(node, FuncExpr))
```
NJrslv, line 1538: I don't quite get why we only consider `FuncExpr` here, what about other types, like window functions or aggregate functions? Why do we even need functions here, why don't just leave `RangeTblEntry`?

gpcontrib/pgaudit/pgaudit.c
```diff
@@ -1501,6 +1502,181 @@ pgaudit_object_access_hook(ObjectAccessType access,
...
+        return parseStr;
+
+    initStringInfo(&str);
+    appendStringInfoString(&str, parseStr);
+    pfree(parseStr);
+
+    foreach(cell, oids)
+    {
+        Oid oid = lfirst_oid(cell);
+        HeapTuple htFunc = SearchSysCache1(PROCOID, ObjectIdGetDatum(oid));
```
NJrslv, line 1684: I am not sure if it is going to hurt the performance, but:
- We are mixing oids in `oids` array, so if some oid is not the `PROCOID`, we will get the cache miss. Is it really that expensive to even think about this? - I dont know. If it is, the easiest solution is to have something like this in order not to cause cache misses:
```c
typedef struct {
    List *rel_oids;
    List *func_oids;
} oids;
```

NJrslv (changes requested):

gpcontrib/pgaudit/pgaudit.c
andr-sokolov: Fixed

gpcontrib/pgaudit/pgaudit.c
andr-sokolov: It is impossible in normal case that att is NULL. If att is NULL, then it is an error. So we should stop

gpcontrib/pgaudit/pgaudit.c
andr-sokolov: The customer want to see mappings for all the functions which oids present in the Query structure

gpcontrib/pgaudit/pgaudit.c
andr-sokolov: Done

gpcontrib/pgaudit/pgaudit.c
andr-sokolov: aggregate and window functions support has been added

NJrslv (approved):

PR #361, opened 2026-03-18
Add the access_log extension

andr-sokolov:
When we need to reduce the used amount of disk space, we should drop some tables
or partitions. In this case, we need to find out tables or partitions which are
not read from. This extension logs when and which user initializes Seq Scan on
which partition or table. The log file name is pg_log/access.log.
Add a new hook - init_scan_hook which is called from InitScanRelation.

### DIFF
--- a/gpcontrib/access_log/access_log.c
+++ b/gpcontrib/access_log/access_log.c
@@ -24,7 +24,8 @@ write_to_log(const char* str)
 	 * the end of the file before writing. The adjustment of the file offset and
 	 * the write operation are performed as an atomic step."
 	 */
-	int f = open(LOG_FILE_NAME, O_WRONLY | O_APPEND | O_CREAT, S_IRUSR | S_IWUSR);
+	int f = OpenTransientFile(LOG_FILE_NAME, O_WRONLY | O_APPEND | O_CREAT,
+							  S_IRUSR | S_IWUSR);
 	if (f < 0 )
 	{
 		ereport(WARNING, (errcode_for_file_access(),
@@ -32,7 +33,7 @@ write_to_log(const char* str)
 		return;
 	}
 	write(f, str, strlen(str));
-	close(f);
+	CloseTransientFile(f);
 }
 
 static void
--- a/gpcontrib/access_log/access_log.c
+++ b/gpcontrib/access_log/access_log.c
@@ -29,7 +29,7 @@ write_to_log(const char* str)
 	if (f < 0 )
 	{
 		ereport(WARNING, (errcode_for_file_access(),
-				errmsg("could not open file " LOG_FILE_NAME)));
+				errmsg("could not open file " LOG_FILE_NAME ": %m")));
 		return;
 	}
 	write(f, str, strlen(str));
--- a/gpcontrib/access_log/access_log.c
+++ b/gpcontrib/access_log/access_log.c
@@ -41,11 +41,13 @@ access_log_init_scan_hook(Relation currentRelation)
 	char		buf[512];
 	HeapTuple	tp;
 	struct timeval	tv;
+	pg_time_t	stamp_time;
 
 	gettimeofday(&tv, NULL);
+	stamp_time = (pg_time_t) tv.tv_sec;
 	pg_strftime(buf, sizeof(buf),
 			"%Y-%m-%d %H:%M:%S        %Z,",
-			pg_localtime((pg_time_t*)&tv.tv_sec, log_timezone));
+			pg_localtime(&stamp_time, log_timezone));
 
 	/* paste milliseconds into place */
 	sprintf(buf + 19, ".%06d", (int) (tv.tv_usec));
--- a/gpcontrib/access_log/access_log.c
+++ b/gpcontrib/access_log/access_log.c
@@ -1,0 +1,91 @@
+#include "postgres.h"
+
+#include "catalog/pg_namespace.h"
+#include "executor/nodeSeqscan.h"
+#include "libpq/auth.h"
+#include "utils/syscache.h"
+
+
+PG_MODULE_MAGIC;
+
+#define LOG_FILE_NAME "pg_log/access.log"
+
+void _PG_init(void);
+
+
+static init_scan_hook_type next_init_scan_hook = NULL;
+
+static void
+write_to_log(const char* str)
+{
+	/*
+	 * Synchronization is not necessary because `man write` says:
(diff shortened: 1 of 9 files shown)


gpcontrib/access_log/access_log.c
```diff
@@ -0,0 +1,91 @@
...
+access_log_init_scan_hook(Relation currentRelation)
+{
+	char		buf[512];
+	HeapTuple	tp;
+	struct timeval	tv;
+
+	gettimeofday(&tv, NULL);
+	pg_strftime(buf, sizeof(buf),
+			"%Y-%m-%d %H:%M:%S        %Z,",
+			pg_localtime((pg_time_t*)&tv.tv_sec, log_timezone));
```
NJrslv: That's a tricky one, I checked that `tv.tv_sec` might be `long int` and `pg_time_t` is always `int64`. In the line `(pg_time_t*)&tv.tv_sec` we might interpret 32-bit variable as a 64-bit one, isn't it an UB?

Check the `elog.c:setup_formatted_log_time(void)` out, they do something similar.

gpcontrib/access_log/access_log.c
```diff
@@ -0,0 +1,91 @@
...
+static void
+write_to_log(const char* str)
+{
+	/*
+	 * Synchronization is not necessary because `man write` says:
+	 * "If the file was open(2)ed with O_APPEND, the file offset is first set to
+	 * the end of the file before writing. The adjustment of the file offset and
+	 * the write operation are performed as an atomic step."
+	 */
+	int f = open(LOG_FILE_NAME, O_WRONLY | O_APPEND | O_CREAT, S_IRUSR | S_IWUSR);
```
NJrslv: A bit general proposition, in PG code the special functions are used that replace sys calls related to files. (Check out `OpenTransientFile()` or `CloseTransientFile()`)
 
These special funcs do something tricky with FDs, I don't quite get it for now, but I think we should use it for the safety purposes.

gpcontrib/access_log/access_log.c
```diff
@@ -0,0 +1,91 @@
...
+	if (f < 0 )
+	{
+		ereport(WARNING, (errcode_for_file_access(),
+				errmsg("could not open file " LOG_FILE_NAME)));
```
NJrslv: I don't think the file name is actually logged :)

gpcontrib/access_log/access_log.c
```diff
@@ -0,0 +1,91 @@
...
+		return;
+	}
+	write(f, str, strlen(str));
+	close(f);
+}
+
+static void
+access_log_init_scan_hook(Relation currentRelation)
+{
+	char		buf[512];
```
NJrslv, line 42: Do you think we can use `StringInfoData` instead of char buf[512]?

The line `buf[19 + 1 + 6] = ' ';` or `sprintf(buf + 19, ".%06d", (int) (tv.tv_usec));` looks quite unhealthy for me.

gpcontrib/access_log/access_log.c
```diff
@@ -0,0 +1,91 @@
...
+
+#include "catalog/pg_namespace.h"
+#include "executor/nodeSeqscan.h"
+#include "libpq/auth.h"
+#include "utils/syscache.h"
+
+
+PG_MODULE_MAGIC;
+
+#define LOG_FILE_NAME "pg_log/access.log"
```
NJrslv, line 11: Are you sure about current dir for the process being datadir? It seems for me that we can be hacked with this, can we use `snprintf(path, MAXPGPATH, "%s/pg_log/access.log", DataDir);`?

gpcontrib/access_log/access_log.c
```diff
@@ -0,0 +1,91 @@
...
+
+void _PG_init(void);
+
+
+static init_scan_hook_type next_init_scan_hook = NULL;
+
+static void
+write_to_log(const char* str)
```
NJrslv, line 19: We call `write_to_log()` func for every scan operation. I think we call this function once per partition scan too.

It does not seem right for me to call open/write/close once per `write_to_log()` invocation, what If a table has a trillion of partitions.

I would open the file lazily and keep a track of its openness, or we can do it in PG_init.

NJrslv (changes requested):

gpcontrib/access_log/access_log.c
andr-sokolov: Using StringInfoData we must use dynamic memory allocation, but using char[] we can avoid it. The code is faster now and there are no potential memory leaks.

The idea is got from `elog.c:setup_formatted_log_time(void)`. We can not translate `tv_usec` to string together with `tv_sec`.  We can use `pg_strftime` to translate `tv_sec` and `sprintf` for `tv_sec`. We cannot combine them in one call.

gpcontrib/access_log/access_log.c
andr-sokolov: Why? When I run `chmod 000 access.log` I see the log record
```
2026-03-25 10:35:58.862485 MSK,"sokolov43a-yu","postgres",p51371,th1289920000,"127.0.0.1","33206",2026-03-25 10:35:40 MSK,0,con8,cmd4,seg0,slice1,,,sx1,"WARNING","42501","could not open file pg_log/access.log",,,,,,,0,,"access_log.c",31,
```

gpcontrib/access_log/access_log.c
andr-sokolov: I'm sure. For example, see `mdunlinkfork`. It calls `do_truncate` and passed `path`. The `path` string is created in `GetRelationPath` and this string in a relative path.

gpcontrib/access_log/access_log.c
andr-sokolov: Someday user will want to rotate the log. They will run something like `mv access.log access.log_20260325`. In this case a new log file should be created when `write_to_log` is called, so we need to open file in `write_to_log`. A partitioned table can have a maximum of 32767 partitions for each partition level (https://docs.arenadata.io/ru/ADB/current/concept/data-model/tables/partitioning.html)

gpcontrib/access_log/access_log.c
andr-sokolov: Fixed

gpcontrib/access_log/access_log.c
andr-sokolov: Fixed

NJrslv (approved):

PR #363, opened 2026-03-23
ORCA: Fix use-after-free in flatten_join_alias_var_optimizer

leborchuk:
Guard pfree/list_free calls with pointer-equality checks to avoid freeing live nodes when flatten_join_alias_vars returns the same pointer unchanged (e.g., outer-reference Vars with varlevelsup > 0).

The unconditional pfree(havingQual) freed the Var node, whose memory was later reused by palloc for a T_List. copyObjectImpl then copied the wrong node type into havingQual, causing ORCA to encounter an unexpected RangeTblEntry and fall back to the Postgres planner.

Applies the same guard pattern to all six fields: targetList, returningList, havingQual, scatterClause, limitOffset, limitCount.

Reported-in: https://github.com/apache/cloudberry/issues/1618

- [ ] Document changes

### DIFF
--- a/src/backend/optimizer/util/clauses.c
+++ b/src/backend/optimizer/util/clauses.c
@@ -5556,35 +5556,45 @@ flatten_join_alias_var_optimizer(Query *query, int queryLevel)
 	if (NIL != targetList)
 	{
 		queryNew->targetList = (List *) flatten_join_alias_vars(root, (Node *) targetList);
-		pfree(targetList);
+		if (targetList != queryNew->targetList)
+			list_free(targetList);
 	}
 
-	List * returningList = queryNew->returningList;
+	List *returningList = queryNew->returningList;
 	if (NIL != returningList)
 	{
 		queryNew->returningList = (List *) flatten_join_alias_vars(root, (Node *) returningList);
-		pfree(returningList);
+		if (returningList != queryNew->returningList)
+			list_free(returningList);
 	}
 
 	Node *havingQual = queryNew->havingQual;
 	if (NULL != havingQual)
 	{
 		queryNew->havingQual = flatten_join_alias_vars(root, havingQual);
-		pfree(havingQual);
+		if (havingQual != queryNew->havingQual)
+		{
+			if (IsA(havingQual, List))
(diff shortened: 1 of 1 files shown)


src/backend/optimizer/util/clauses.c
```diff
@@ -5556,35 +5556,40 @@ flatten_join_alias_var_optimizer(Query *query, int queryLevel)
...
+		if (returningList != queryNew->returningList)
+			list_free(returningList);
 	}
 
 	Node *havingQual = queryNew->havingQual;
 	if (NULL != havingQual)
 	{
 		queryNew->havingQual = flatten_join_alias_vars(root, havingQual);
-		pfree(havingQual);
+		if (havingQual != queryNew->havingQual)
```
NJrslv, line 5575: In the original patch, they have such a code:
```C
		if (havingQual != queryNew->havingQual)
		{
			if (IsA(havingQual, List))
				list_free((List *) havingQual);
			else
				pfree(havingQual);
		}
```
Our version has:
```C
		if (havingQual != queryNew->havingQual)
			pfree(havingQual);
```

Is there any reason why the extra if was removed?

leborchuk: Yep, you are right. This fix is from the 2nd commit in https://github.com/apache/cloudberry/pull/1620/

I took only the 1st commit  [ORCA: Fix use-after-free in flatten_join_alias_var_optimizer](https://github.com/apache/cloudberry/pull/1620/changes/f07c7c359fb3fd267ca2c4c4ac1ecf10d04eec0c)

But intentionally ommiited the 2nd commit [ORCA: Fix incorrect decorrelation of GROUP BY () HAVING <outer_ref>](https://github.com/apache/cloudberry/pull/1620/changes/7983b7a3b37ffeac36dbec4affc95bf61d783339)

Since it commit depends on other commits adding groupingsets to GPORCA, we do not want a massive refactoring in GP6. 

Added also for checking if havingQual is a list - make sense to copy it here too

NJrslv (approved):

PR #370, opened 2026-04-06
Fix cross-slice Shared Scan failure for CTE with replicated table in scalar subquery

Alena0704:
Fix cross-slice SharedScan hang for replicated CTE in scalar subquery

When a CTE over a replicated table is referenced from multiple scalar
subqueries, ORCA puts the CTE Consumer inside a SubPlan that runs on a
different slice than the Producer. The cross-slice temp file protocol
does not work for this topology, so the query either hangs or fails
with "could not open existing temporary file".

Fix in apply_shareinput_xslice pass 2: when a cross-slice Consumer is
found inside a SubPlan and the CTE source is a replicated table, give
the Consumer its own local copy of the Producer's subtree (Materialize
and base Scan) with a fresh share_id and SHARE_MATERIAL. Since the table
is replicated, every segment already has the full data, so a local
copy is equivalent and removes the cross-slice coordination. Other
Consumers of the same CTE in the same slice reuse this inlined copy.

### DIFF
--- a/src/backend/cdb/cdbmutate.c
+++ b/src/backend/cdb/cdbmutate.c
@@ -2494,16 +2494,12 @@ shareinput_mutator_xslice_2(Node *node, PlannerInfo *root, bool fPop)
 			if (shareSliceId != motId)
 			{
 				/*
-				 * Check for cross-slice SharedScan with a replicated
-				 * table inside a SubPlan.  When the producer runs on
-				 * fewer segments than the consumer, the temp file does
-				 * not exist on all consumer segments.
-				 *
-				 * Fix: give this consumer its own copy of the
-				 * producer's underlying plan (Materialize + Scan) and
-				 * convert it to an intra-slice SHARE_MATERIAL with a
-				 * new share_id.  The consumer then materializes data
-				 * locally instead of reading cross-slice temp files.
+				 * Cross-slice SharedScan over a replicated table inside
+				 * a SubPlan: the cross-slice temp file protocol does not
+				 * work in this topology.  Give this Consumer a local copy
+				 * of the Producer's subtree with a fresh share_id /
+				 * SHARE_MATERIAL -- replicated data is identical on every
+				 * segment, so a local copy is equivalent.
 				 */
 				bool		inlined = false;
 
@@ -2513,12 +2509,7 @@ shareinput_mutator_xslice_2(Node *node, PlannerInfo *root, bool fPop)
 					int			existingNewId = -1;
 					int			k;
 
-					/*
-					 * Check if we already inlined a consumer for this
-					 * same original share_id in this same slice.  If so,
-					 * make this consumer a reader of the existing inlined
-					 * producer instead of creating another copy.
-					 */
+					/* Already inlined in this slice: reuse it as an intra-slice reader. */
 					for (k = 0; k < ctxt->inlined_count; k++)
 					{
 						if (ctxt->inlined_orig_ids[k] == origShareId &&
@@ -2531,12 +2522,7 @@ shareinput_mutator_xslice_2(Node *node, PlannerInfo *root, bool fPop)
 
 					if (existingNewId >= 0)
 					{
-						/*
-						 * Reuse the already-inlined producer: make this
-						 * consumer a reader with the same share_id.
-						 * This is intra-slice sharing, so no need to
-						 * increment nsharer_xslice.
-						 */
+						/* Intra-slice sharing -- do not bump nsharer_xslice. */
 						if (origShareId < ctxt->orig_producer_count)
 							ctxt->consumer_counts[origShareId]--;
 
@@ -2555,12 +2541,7 @@ shareinput_mutator_xslice_2(Node *node, PlannerInfo *root, bool fPop)
 						while (leaf && leaf->lefttree)
 							leaf = leaf->lefttree;
 
-						/*
-						 * Any scan over a base relation shares the Scan
-						 * struct layout (scanrelid at a fixed offset):
-						 * SeqScan, IndexScan, IndexOnlyScan, BitmapHeapScan,
-						 * TidScan, SampleScan, and the Dynamic* variants.
-						 */
+						/* Base scan over a replicated table -- safe to inline locally. */
 						if (leaf &&
 							(IsA(leaf, SeqScan) ||
 							 IsA(leaf, IndexScan) ||
@@ -2592,11 +2573,7 @@ shareinput_mutator_xslice_2(Node *node, PlannerInfo *root, bool fPop)
 
 										pfree(policy);
 
-										/*
-										 * Deep copy the producer's subtree and
-										 * assign a fresh share_id so there is no
-										 * conflict with the original producer.
-										 */
+										/* Deep-copy with a fresh share_id. */
 										if (origShareId < ctxt->orig_producer_count)
 											ctxt->consumer_counts[origShareId]--;
 
@@ -2610,10 +2587,7 @@ shareinput_mutator_xslice_2(Node *node, PlannerInfo *root, bool fPop)
 										sisc->driver_slice = motId;
 										plan->lefttree = newChild;
 
-										/*
-										 * Register the new producer so later
-										 * passes can find it.
-										 */
+										/* Register the new producer for later passes. */
 										ctxt->producers = repalloc(ctxt->producers,
 											ctxt->producer_count * sizeof(ShareInputScan *));
 										ctxt->producers[newShareId] = sisc;
@@ -2621,11 +2595,7 @@ shareinput_mutator_xslice_2(Node *node, PlannerInfo *root, bool fPop)
 											ctxt->producer_count * sizeof(int));
 										ctxt->sliceMarks[newShareId] = motId;
 
-										/*
-										 * Record the mapping so subsequent consumers
-										 * of the same CTE in this slice can reuse
-										 * this inlined producer.
-										 */
+										/* Let later Consumers of the same CTE in this slice reuse it. */
 										ctxt->inlined_count++;
 										ctxt->inlined_orig_ids = repalloc(ctxt->inlined_orig_ids,
 											ctxt->inlined_count * sizeof(int));
@@ -2775,10 +2745,9 @@ shareinput_mutator_xslice_4(Node *node, PlannerInfo *root, bool fPop)
 }
 
 /*
- * record_subplan_motid_walker
- *   Walk expressions looking for SubPlan references.  For each SubPlan found,
- *   record the current motId from the motion stack.  This tells us what slice
- *   the SubPlan actually executes in (which is the caller's slice).
+ * For each SubPlan expression, record the motId of the slice that calls
+ * it; later we use this to tell whether a Consumer lives inside a SubPlan
+ * slice.
  */
 static bool
 record_subplan_motid_walker(Node *node, ApplyShareInputContext *ctxt)
--- a/src/backend/cdb/cdbmutate.c
+++ b/src/backend/cdb/cdbmutate.c
@@ -2437,6 +2437,12 @@ shareinput_mutator_xslice_1(Node *node, PlannerInfo *root, bool fPop)
 			ctxt->producers[sisc->share_id] = sisc;
 			ctxt->sliceMarks[sisc->share_id] = motId;
 		}
+		else
+		{
+			/* Consumer: count references to original producers */
+			if (sisc->share_id < ctxt->orig_producer_count)
+				ctxt->consumer_counts[sisc->share_id]++;
+		}
 	}
 
 	return true;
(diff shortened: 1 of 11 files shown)


src/test/regress/sql/shared_scan.sql
```diff
@@ -137,3 +137,42 @@ select case when ten < 5 then ten else ten * 2 end, count(distinct two), count(d
...
+-- consumers (SubPlans), causing "could not open existing temporary file" error
+-- or a hang.  The fix ensures:
+-- 1. ORCA places both producer and consumer in the same slice (via a 3rd
+--    distribution request in CPhysicalSequence).
+-- 2. apply_shareinput_xslice correctly determines the execution slice for
+--    SharedScan consumers inside SubPlans.
+SET search_path = shared_scan;
+
+CREATE TABLE ss_t1 (id int) DISTRIBUTED BY (id);
+INSERT INTO ss_t1 SELECT i FROM generate_series(1, 40000) i;
```
NJrslv: Do we actually need 40k rows here? It sounds like an extra load on CI that can be avoided (not sure though).

src/test/regress/sql/shared_scan.sql
```diff
@@ -137,3 +137,42 @@ select case when ten < 5 then ten else ten * 2 end, count(distinct two), count(d
...
+INSERT INTO ss_t2 VALUES (1, 10), (2, 20);
+ANALYZE ss_t1;
+ANALYZE ss_t2;
+
+SET statement_timeout = '15s';
+
+WITH
+  cte1 AS (SELECT v FROM ss_t2 WHERE id = 1),
+  cte2 AS (SELECT v FROM ss_t2 WHERE id = 2)
+SELECT DISTINCT
```
NJrslv: I think we need to regenerate expected/*.out. The test is failing right now, here is the diff:
```diff
--- \/__w\/gpdb\/gpdb\/src\/test\/regress\/expected\/shared_scan\.out	2026-04-06 12:06:49.948572073 +0100
+++ \/__w\/gpdb\/gpdb\/src\/test\/regress\/results\/shared_scan\.out	2026-04-06 12:06:49.954572069 +0100
@@ -285,6 +286,18 @@
 ANALYZE ss_t1;
 ANALYZE ss_t2;
 SET statement_timeout = '15s';
+WITH
+  cte1 AS (SELECT v FROM ss_t2 WHERE id = 1),
+  cte2 AS (SELECT v FROM ss_t2 WHERE id = 2)
+SELECT DISTINCT
+       (SELECT v FROM cte1) + (SELECT v FROM cte2) +
+       (SELECT v FROM cte1) + (SELECT v FROM cte2) AS result
+FROM ss_t1;
+ result 
+--------
+     60
+(1 row)
+
 -- Must not error with "could not open existing temporary file" or hang.
 SELECT count(*) FROM (
   WITH
@@ -294,7 +307,7 @@
          (SELECT v FROM cte1) + (SELECT v FROM cte2) AS result
   FROM ss_t1
 ) q WHERE result = 60;
- count
+ count 
 -------
  40000
 (1 row)

======================================================================

```

src/backend/cdb/cdbmutate.c
```diff
@@ -2611,12 +2611,107 @@ shareinput_mutator_xslice_4(Node *node, PlannerInfo *root, bool fPop)
...
+		Motion	   *motion = (Motion *) plan;
+
+		shareinput_pushmot(ctxt, motion->motionID);
+		return true;
+	}
+
+	/* Scan expressions for SubPlan references */
+	record_subplan_motid_walker((Node *) plan->targetlist, ctxt);
+	record_subplan_motid_walker((Node *) plan->qual, ctxt);
+
```
NJrslv: Check out this function and specifically everything around this line:
https://github.com/open-gpdb/gpdb/blob/a89823aefcd2a4e055b34507656dc10cc7b2c66e/src/backend/optimizer/plan/subselect.c#L2514

I am not so sure we handle every case here, like in the function above, what if SubPlan lands in joinqual, not in targetlist or qual? Is it even possible?

Don't we want something like this?
```c
record_subplan_motid_walker((Node *) plan->targetlist, ctxt);
record_subplan_motid_walker((Node *) plan->qual, ctxt);

switch (nodeTag(plan))
{
    case T_Result:
        record_subplan_motid_walker(((Result *) plan)->resconstantqual, ctxt);
        break;
    case T_IndexScan:
        record_subplan_motid_walker((Node *) ((IndexScan *) plan)->indexqual, ctxt);
        record_subplan_motid_walker((Node *) ((IndexScan *) plan)->indexorderby, ctxt);
...
```

Also, a good idea to submit this PR to the Cloudberry folks, so they can participate in the review process.

NJrslv (changes requested):

src/test/regress/sql/shared_scan.sql
Alena0704: Yes, we can cut them down to 100 or less without any problems

src/test/regress/sql/shared_scan.sql
Alena0704: Agree

src/backend/cdb/cdbmutate.c
Alena0704: > > I am not so sure we handle every case here, like in the function above, what if SubPlan lands in joinqual, not in targetlist or qual? Is it even possible?

Yes, you are right here - for exampe, select ... from a join b on a.x = (select ...)).
I have added the recursive pass for different types of node because elements about identificator of subplan can be there. now all three join types and others that can contain specific fields walk quals for SubPlan/Param references.

src/test/regress/sql/shared_scan.sql
Alena0704: Fixed

src/test/regress/sql/shared_scan.sql
Alena0704: Fixed - I used 10 :)

src/backend/cdb/cdbmutate.c
Alena0704: Thank you for your review!

NJrslv:
@Alena0704, weirdness alert, CI tests with the optimizer on are executing for more than 120 mins, so they are being cancelled. Why? - IDK.

I'd suggest making sure we introduce no regressions and pinging +1/+2 folks to eyeball the PR. Thanks!

We also have a diff in gporca regress tests:
```diff
--- \/__w\/gpdb\/gpdb\/src\/test\/regress\/expected\/gporca_optimizer\.out	2026-04-09 08:49:26.330139072 +0100
+++ \/__w\/gpdb\/gpdb\/src\/test\/regress\/results\/gporca\.out	2026-04-09 08:49:26.617144383 +0100
@@ -15254,36 +15288,39 @@
 from mat m1 join mat m2 on m1.j = m2.j;
                                                                                       QUERY PLAN                                                                                       
 ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+                                       Output: share0_ref2.i, share0_ref2.j
+                                 ->  Shared Scan (share slice:id 1:0)  (cost=0.00..431.00 rows=1 width=4)
+                                 Filter: (share0_ref2.j = 2)
                                  Index Cond: (material_bitmapscan.k = 'Thu Jun 03 00:00:00 2021'::timestamp without time zone)
-                                 Output: share0_ref2.i, share0_ref2.j
+                                 Output: share0_ref2.j
+                                 Output: share0_ref3.i, share0_ref3.j
                            ->  Bitmap Index Scan on material_bitmapscan_idx  (cost=0.00..0.00 rows=0 width=0)
-                           ->  Shared Scan (share slice:id 1:0)  (cost=0.00..431.00 rows=1 width=4)
+                           ->  Result  (cost=0.00..431.00 rows=1 width=4)
+                           ->  Shared Scan (share slice:id 1:0)  (cost=0.00..431.00 rows=1 width=8)
                            Filter: ((material_bitmapscan.i = 2) AND (material_bitmapscan.j = 2) AND (material_bitmapscan.l = 'Thu Jun 03 00:00:00 2021'::timestamp without time zone))
-                           Filter: (share0_ref2.j = 2)
+                           Filter: (share0_ref3.j = 2)
                            Output: material_bitmapscan.i, material_bitmapscan.j
                            Output: share0_ref2.j
                            Output: share0_ref3.i, share0_ref3.j
                            Recheck Cond: (material_bitmapscan.k = 'Thu Jun 03 00:00:00 2021'::timestamp without time zone)
                      ->  Bitmap Heap Scan on orca.material_bitmapscan  (cost=0.00..390.25 rows=1 width=8)
-                     ->  Result  (cost=0.00..431.00 rows=1 width=4)
-                     ->  Shared Scan (share slice:id 1:0)  (cost=0.00..431.00 rows=1 width=8)
-                     Filter: (share0_ref3.j = 2)
+                     ->  Hash  (cost=431.00..431.00 rows=1 width=4)
+                     ->  Result  (cost=0.00..431.00 rows=1 width=8)
+                     Hash Cond: (share0_ref3.j = share0_ref2.j)
                      Output: material_bitmapscan.i, material_bitmapscan.j
-                     Output: share0_ref2.j
-                     Output: share0_ref3.i, share0_ref3.j
-               ->  Hash  (cost=431.00..431.00 rows=1 width=4)
+                     Output: share0_ref3.i
+               ->  Hash Join  (cost=0.00..862.00 rows=1 width=4)
                ->  Materialize  (cost=0.00..390.25 rows=1 width=1)
-               ->  Result  (cost=0.00..431.00 rows=1 width=8)
-               Hash Cond: (share0_ref3.j = share0_ref2.j)
+               One-Time Filter: (gp_execution_segment() = ###)
                Output: share0_ref1.i, share0_ref1.j
                Output: share0_ref3.i
-         ->  Hash Join  (cost=0.00..862.00 rows=1 width=4)
+         ->  Result  (cost=0.00..862.00 rows=1 width=4)
          ->  Shared Scan (share slice:id 1:0)  (cost=0.00..390.25 rows=1 width=1)
          Output: share0_ref3.i
    ->  Sequence  (cost=0.00..1252.25 rows=1 width=4)
    Output: share0_ref3.i
- Gather Motion 1:1  (slice1; segments: 1)  (cost=0.00..1252.25 rows=1 width=4)
-(31 rows)
+ Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..1252.25 rows=1 width=4)
+(34 rows)
 
 with mat as(
     select i, j from material_bitmapscan where i = 2 and j = 2

======================================================================

```

Alena0704:
It is not weird. 
I ran into an issue with the first solution: it’s hard to distinguish a CTE Sequence from a percentile Sequence.

The problem is that if we unconditionally assign Any to child 1 for all replicated Sequences, it breaks percentile aggregates (they also use Sequence with a CTE producer, and Any allows ORCA to pick a trivial plan without a Shared Scan). The tests confirmed this. Also, for non-replicated CTEs this causes:
CTE Producer–Consumer execution locality mismatch (producer on master, consumer on segments → fallback to planner).

I couldn’t find a clean way in PdsRequired to distinguish CTE + replicated from other Sequences.

So I decided to try fixing it in the second rule:

Child 0 gets Any → CTE producer scans a replicated table → produces Replicated
The code sees Replicated → requests Replicated for child 1
Both children are Replicated → Sequence is Replicated → Gather Motion 1:1 (single segment)

There is an optimization here: instead of running the Sequence on all 3 segments (where each does the same work on replicated data), we run it on just one. Less work, one segment instead of three.

This optimization is exactly what causes the problem:
Sequence runs in slice 2 on 1 segment, while SubPlans run in slice 1 on 3 segments → the temp file exists only on 1 segment → crash.

I disabled this by assigning NonSingleton instead of Replicated to the second child. This forces the Sequence to run on all segments (Gather Motion 3:1), keeping CTE producers and consumers in the same slice.

So far, the only side effect is the material_bitmapscan test (plan changes from 1:1 to 3:1) and the tests that were previously timing out.

After seeing this, I actually went back to the first approach again - trying to fix this more precisely for this specific type of query. If that doesn’t work, I think it makes sense to add a fix with a fallback for the first optimization.

Alena0704:
I'm trying this: when the first child Sequence delivers Replicated, we request NonSingleton on the second (instead of Replicated as in the second request) – like before https://github.com/open-gpdb/gpdb/commit/51b4ddbcbe0550374456a46ca15dcff411d29b5b, only we're trying to do this selectively and adding it as an alternative. This solves our original problem – ORCA can choose the Gather 3:1 plan and shouldn't break current requests.
I'm waiting for the CI tests to pass.

Alena0704:
I fixed it differently.
When a cross-slice consumer in SubPlan references a replicated table, instead of marking it as XSLICE (which requires a temp file on all segments):

I create a copyObject producer's lefttree (Materialize + SeqScan)
Assigned a new share_id (so it doesn't conflict with the original producer)
Set share_type = SHARE_MATERIAL (intra-slice, without temp files)
Attached the copy as a lefttree consumer
Registered in ctxt->producers[]
The consumer becomes an independent producer and runs Materialize + SeqScan locally on each segment. The replicated data is the same on all segments, so everything is correct.

Alena0704:
During review, Andrei have noticed another problem - redundant producers and consumers have been created:
cte1 and cte2 are each mentioned twice → 4 consumers → 4 independent copies (share 1:2, 1:3, 1:4, 1:5), although SubPlans 1 and 3 scan the same thing. 
````
postgres=# explain analyze with
  cte1 as ( select v from t2 where id = 1 ),
  cte2 as ( select v from t2 where id = 2 )
select (select v from cte1) + (select v from cte2) + (select v from cte1) + (select v from cte2)
from t1;
                                                                                QUERY PLAN                                                                        
         
------------------------------------------------------------------------------------------------------------------------------------------------------------------
---------
 Gather Motion 1:1  (slice2; segments: 1)  (cost=0.00..1422945480902069.75 rows=40000 width=4) (actual time=37.957..900.913 rows=40000 loops=1)
   ->  Sequence  (cost=0.00..1422945480902068.00 rows=40000 width=4) (actual time=22.540..883.504 rows=40000 loops=1)
         ->  Shared Scan (share slice:id 2:0)  (cost=0.00..431.00 rows=1 width=1) (never executed)
               ->  Materialize  (cost=0.00..431.00 rows=1 width=1) (never executed)
                     ->  Seq Scan on t2 t2_1  (cost=0.00..431.00 rows=1 width=4) (never executed)
                           Filter: (id = 1)
         ->  Sequence  (cost=0.00..1422945480901636.75 rows=40000 width=4) (actual time=22.503..645.695 rows=40000 loops=1)
               ->  Shared Scan (share slice:id 2:1)  (cost=0.00..431.00 rows=1 width=1) (never executed)
                     ->  Materialize  (cost=0.00..431.00 rows=1 width=1) (never executed)
                           ->  Seq Scan on t2  (cost=0.00..431.00 rows=1 width=4) (never executed)
                                 Filter: (id = 2)
               ->  Result  (cost=0.00..1422945480901205.50 rows=40000 width=4) (actual time=22.473..405.689 rows=40000 loops=1)
                     ->  Broadcast Motion 3:1  (slice1; segments: 3)  (cost=0.00..1422945480901205.25 rows=1000 width=16) (actual time=22.445..159.638 rows=40000 
loops=1)
                           ->  Result  (cost=0.00..1422945480901205.00 rows=334 width=16) (actual time=0.238..386.993 rows=13426 loops=1)
                                 ->  Seq Scan on t1  (cost=0.00..1324358.39 rows=334 width=4) (actual time=0.113..36.145 rows=13426 loops=1)
                                 SubPlan 1  (slice1; segments: 3)
                                   ->  Shared Scan (share slice:id 1:2)  (cost=0.00..431.00 rows=1 width=4) (never executed)
                                         ->  Materialize  (cost=0.00..431.00 rows=1 width=1) (never executed)
                                               ->  Seq Scan on t2 t2_1  (cost=0.00..431.00 rows=1 width=4) (never executed)
                                                     Filter: (id = 1)
                                 SubPlan 2  (slice1; segments: 3)
                                   ->  Shared Scan (share slice:id 1:3)  (cost=0.00..431.00 rows=1 width=4) (never executed)
                                         ->  Materialize  (cost=0.00..431.00 rows=1 width=1) (never executed)
                                               ->  Seq Scan on t2  (cost=0.00..431.00 rows=1 width=4) (never executed)
                                                     Filter: (id = 2)
                                 SubPlan 3  (slice1; segments: 3)
                                   ->  Shared Scan (share slice:id 1:4)  (cost=0.00..431.00 rows=1 width=4) (never executed)
                                         ->  Materialize  (cost=0.00..431.00 rows=1 width=1) (never executed)
                                               ->  Seq Scan on t2 t2_1  (cost=0.00..431.00 rows=1 width=4) (never executed)
                                                     Filter: (id = 1)
                                 SubPlan 4  (slice1; segments: 3)
                                   ->  Shared Scan (share slice:id 1:5)  (cost=0.00..431.00 rows=1 width=4) (never executed)
                                         ->  Materialize  (cost=0.00..431.00 rows=1 width=1) (never executed)
                                               ->  Seq Scan on t2  (cost=0.00..431.00 rows=1 width=4) (never executed)
                                                     Filter: (id = 2)
 Planning time: 194.545 ms
   (slice0)    Executor memory: 375K bytes.
   (slice1)    Executor memory: 268K bytes avg x 3 workers, 268K bytes max (seg0).
   (slice2)    Executor memory: 312K bytes (seg2).
 Memory used:  128000kB
 Optimizer: Pivotal Optimizer (GPORCA)
 Execution time: 1019.032 ms
(42 rows)
```
`
I have fixed it in the following commit.I have tracked (orig_share_id, motId) and created new_share_id. When re-inlining the same CTE in the same slice, the consumer becomes a reader of the already created producer. Result: 2 producers + 2 readers instead of 4 producers. I counted consumers in pass 1, decrement when inlining in pass 2
After 4 passes, cleanup_orphaned_producers() removes SharedScans with zero consumers from the Sequence
Nested Sequences (remaining after removal) are flattened into the parent


query plan looks like this now:
`postgres=# explain analyze with
  cte1 as ( select v from t2 where id = 1 ),
  cte2 as ( select v from t2 where id = 2 )
select (select v from cte1) + (select v from cte2) + (select v from cte1) + (select v from cte2)
from t1;
                                                                            QUERY PLAN                                                                             
-------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Gather Motion 1:1  (slice2; segments: 1)  (cost=0.00..1422945480902069.75 rows=40000 width=4) (actual time=3.920..59.581 rows=40000 loops=1)
   ->  Sequence  (cost=0.00..1422945480902068.00 rows=40000 width=4) (actual time=2.004..45.390 rows=40000 loops=1)
         ->  Result  (cost=0.00..1422945480901205.50 rows=40000 width=4) (actual time=2.004..40.522 rows=40000 loops=1)
               ->  Broadcast Motion 3:1  (slice1; segments: 3)  (cost=0.00..1422945480901205.25 rows=1000 width=16) (actual time=1.991..18.081 rows=40000 loops=1)
                     ->  Result  (cost=0.00..1422945480901205.00 rows=334 width=16) (actual time=0.081..22.256 rows=13426 loops=1)
                           ->  Seq Scan on t1  (cost=0.00..1324358.39 rows=334 width=4) (actual time=0.067..4.317 rows=13426 loops=1)
                           SubPlan 1  (slice1; segments: 3)
                             ->  Shared Scan (share slice:id 1:2)  (cost=0.00..431.00 rows=1 width=4) (never executed)
                                   ->  Materialize  (cost=0.00..431.00 rows=1 width=1) (never executed)
                                         ->  Seq Scan on t2 t2_1  (cost=0.00..431.00 rows=1 width=4) (never executed)
                                               Filter: (id = 1)
                           SubPlan 2  (slice1; segments: 3)
                             ->  Shared Scan (share slice:id 1:3)  (cost=0.00..431.00 rows=1 width=4) (never executed)
                                   ->  Materialize  (cost=0.00..431.00 rows=1 width=1) (never executed)
                                         ->  Seq Scan on t2  (cost=0.00..431.00 rows=1 width=4) (never executed)
                                               Filter: (id = 2)
                           SubPlan 3  (slice1; segments: 3)
                             ->  Shared Scan (share slice:id 1:2)  (cost=0.00..431.00 rows=1 width=4) (never executed)
                           SubPlan 4  (slice1; segments: 3)
                             ->  Shared Scan (share slice:id 1:3)  (cost=0.00..431.00 rows=1 width=4) (never executed)
 Planning time: 177.929 ms
   (slice0)    Executor memory: 247K bytes.
   (slice1)    Executor memory: 204K bytes avg x 3 workers, 204K bytes max (seg0).
   (slice2)    Executor memory: 151K bytes (seg1).
 Memory used:  128000kB
 Optimizer: Pivotal Optimizer (GPORCA)
 Execution time: 84.155 ms
(27 rows)`

PR #372, opened 2026-04-07
Add custom nodes gpdb

Alena0704:
Port CustomScan and ExtensibleNode infrastructure from PostgreSQL to Greenplum
    
    Add support for custom scan nodes, allowing extensions to define
    their own scan strategies that integrate with the planner, executor,
    EXPLAIN, and GPDB's CDB dispatch pipeline.
    
    New node types: T_CustomScan, T_CustomScanState, T_CustomPath,
    T_ExtensibleNode. New files: extensible.h/c (registration framework),
    nodeCustom.h/c (executor callbacks). Full binary serialization support
    for plan dispatch to segments. Includes set_rel_pathlist_hook for
    planner extensibility and a test extension (custom_scan_test) that
    replaces SeqScan with a CustomScan to validate the pipeline end-to-end.

custom_scan_test extension is an example how to use it and simple example to check that it works


NJrslv:
I think we need to update the test, the lines have shifted a bit, please see the regress diff down below:
```diff
--- \/__w\/gpdb\/gpdb\/src\/test\/regress\/expected\/join_gp\.out	2026-04-07 16:05:41.776158435 +0100
+++ \/__w\/gpdb\/gpdb\/src\/test\/regress\/results\/join_gp\.out	2026-04-07 16:05:41.824158392 +0100
@@ -1772,7 +1772,7 @@
           (box(point(0.8,0.8), point(1.0,1.0)))) as v(bb)
 cross join lateral
   (select p from gist_tbl_github9733 where p <@ bb order by p <-> bb[0] limit 2) ss;
-ERROR:  could not devise a query plan for the given query (pathnode.c:422)
+ERROR:  could not devise a query plan for the given query (pathnode.c:424)
 reset enable_seqscan;
 explain (costs off)
 select p from

======================================================================
```

src/backend/executor/nodeCustom.c
```diff
@@ -0,0 +1,170 @@
...
+					 (PlanState *) css);
+
+	/*
+	 * For projections, if the custom scan tlist is set, we project from
+	 * INDEX_VAR; otherwise use the scan relation.
+	 */
+	if (tlistvarno == INDEX_VAR)
+		ExecAssignScanProjectionInfo(&css->ss);
+	else
+		ExecAssignScanProjectionInfo(&css->ss);
```
NJrslv, line 105: What about two identical branches? I also don't see this code in the upstream, can you please point to the code you are porting from?

I see that upstream has different code in some places but I don't know if it's okay or not.

NJrslv (changes requested):

src/backend/executor/nodeCustom.c
Alena0704: It is from Postgresql code master branch
I tried to move custom nodes in gpdb because some features requires their own algorithms or path preparation logic.

PR #375, opened 2026-04-13
orca: fallback to Postgres optimizer on cross-slice replicated CTE Consumer

Alena0704:
orca: fallback to Postgres optimizer on cross-slice replicated CTE Consumer
Inspired by greengage https://github.com/open-gpdb/gpdb/commit/51fe92e04b5b22fb465117cfc7dd8c2517f40cc2: before Expr->DXL translation,
walk the physical tree and track which slice each CTE Producer
and Consumer lives on. If a Consumer is on a different slice
than its Producer and the Producer's distribution is replicated,
force a fallback to the Postgres optimizer.

The replicated filter is essential: ordinary cross-slice CTE plans
(non-replicated Producer with Gather/Redistribute Consumer) are a
normal ORCA pattern and must not trigger fallback.

https://github.com/open-gpdb/gpdb/commit/51fe92e04b5b22fb465117cfc7dd8c2517f40cc2 doesn't trigger when a CTE over a replicated table is
referenced from a scalar subquery, so the query hangs. This commit
replaces the single-point check with a whole-tree walker that
catches both cases.

Tests: bring the original qp_orca_fallback case from https://github.com/open-gpdb/gpdb/commit/51fe92e04b5b22fb465117cfc7dd8c2517f40cc2 and add
a scalar-subquery reproducer in shared_scan guarded by
statement_timeout.

### DIFF
--- a/src/backend/gporca/libgpopt/src/base/CUtils.cpp
+++ b/src/backend/gporca/libgpopt/src/base/CUtils.cpp
@@ -901,6 +901,135 @@ CUtils::FHasCTEAnchor(CExpression *pexpr)
 	return false;
 }
 
+// True if the distribution is replicated-like.
+static BOOL
+FReplicatedLikeDistribution(CDistributionSpec::EDistributionType edt)
+{
+	return (CDistributionSpec::EdtStrictReplicated == edt ||
+			CDistributionSpec::EdtTaintedReplicated == edt ||
+			CDistributionSpec::EdtUniversal == edt);
+}
+
+struct SCTEProducerInfo
+{
+	ULONG cteId;
+	ULONG sliceId;
+	BOOL isReplicated;
+};
+
+typedef CDynamicPtrArray<SCTEProducerInfo, CleanupDelete<SCTEProducerInfo> >
+	CTEProducerInfoArray;
+
+// Walk the physical tree, recording the slice id of every Physical CTE
+// Producer and Consumer. For each Producer we also remember whether
+// its (child's) distribution is replicated-like. Slices are delimited
+// by Motion nodes: each non-scalar child of a Motion lives in a fresh
+// slice -- same motId-stack idea as in apply_shareinput_xslice.
+static void
+CollectCTESlices(CMemoryPool *mp, CExpression *pexpr, ULONG curSlice,
+				 ULONG *pNextSlice, CTEProducerInfoArray *prodInfos,
+				 ULongPtrArray *consIds, ULongPtrArray *consSlices)
+{
+	COperator *pop = pexpr->Pop();
+
+	if (COperator::EopPhysicalCTEProducer == pop->Eopid())
+	{
+		// Producer's distribution comes from its only child -- inspect
+		// it there.
+		GPOS_ASSERT(1 == pexpr->Arity());
+		CExpression *pexprChild = (*pexpr)[0];
+		CDrvdPropPlan *pdpplan =
+			CDrvdPropPlan::Pdpplan(pexprChild->PdpDerive());
+
+		SCTEProducerInfo *info = GPOS_NEW(mp) SCTEProducerInfo;
+		info->cteId = CPhysicalCTEProducer::PopConvert(pop)->UlCTEId();
+		info->sliceId = curSlice;
+		info->isReplicated =
+			FReplicatedLikeDistribution(pdpplan->Pds()->Edt());
+		prodInfos->Append(info);
+	}
+	else if (COperator::EopPhysicalCTEConsumer == pop->Eopid())
+	{
+		// Consumer is a leaf -- only record (cteId, curSlice). The
+		// cross-slice decision is made by the caller after the whole
+		// tree has been walked.
+		consIds->Append(GPOS_NEW(mp) ULONG(
+			CPhysicalCTEConsumer::PopConvert(pop)->UlCTEId()));
+		consSlices->Append(GPOS_NEW(mp) ULONG(curSlice));
+	}
+
+	BOOL isMotion = CUtils::FPhysicalMotion(pop);
+
+	for (ULONG ul = 0; ul < pexpr->Arity(); ul++)
+	{
+		CExpression *pexprChild = (*pexpr)[ul];
+
+		// Scalar subtrees (predicates, project elements) never run as
+		// separate executor groups, so they cannot host a slice.
+		if (pexprChild->Pop()->FScalar())
+		{
+			continue;
+		}
+
+		// Allocate a fresh slice id for each non-scalar child of a
+		// Motion; otherwise the child stays in the parent's slice.
+		ULONG childSlice = curSlice;
+		if (isMotion)
+		{
+			(*pNextSlice)++;
+			childSlice = *pNextSlice;
+		}
+
+		CollectCTESlices(mp, pexprChild, childSlice, pNextSlice, prodInfos,
+						 consIds, consSlices);
+	}
+}
+
+BOOL
+CUtils::FHasCrossSliceReplicatedCTEConsumer(CMemoryPool *mp, CExpression *pexpr)
+{
+	if (NULL == pexpr)
+	{
+		return false;
+	}
+
+	CTEProducerInfoArray *prodInfos = GPOS_NEW(mp) CTEProducerInfoArray(mp);
+	ULongPtrArray *consIds = GPOS_NEW(mp) ULongPtrArray(mp);
+	ULongPtrArray *consSlices = GPOS_NEW(mp) ULongPtrArray(mp);
+	ULONG nextSlice = 0;
+
+	CollectCTESlices(mp, pexpr, 0 /*curSlice*/, &nextSlice, prodInfos,
+					 consIds, consSlices);
+
+	BOOL cross = false;
+
+	for (ULONG ic = 0; ic < consIds->Size() && !cross; ic++)
+	{
+		ULONG cid = *(*consIds)[ic];
+		ULONG cslice = *(*consSlices)[ic];
+
+		for (ULONG ip = 0; ip < prodInfos->Size(); ip++)
+		{
+			SCTEProducerInfo *info = (*prodInfos)[ip];
+			if (info->cteId != cid)
+			{
+				continue;
+			}
+			if (info->isReplicated && info->sliceId != cslice)
+			{
+				cross = true;
+			}
+			break;
+		}
+	}
+
+	prodInfos->Release();
+	consIds->Release();
+	consSlices->Release();
+
+	return cross;
+}
+
 //---------------------------------------------------------------------------
 //	@class:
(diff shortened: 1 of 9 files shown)


andr-sokolov (changes requested):
Test doesn't reproduce the error without the patch

src/test/regress/sql/shared_scan.sql
```diff
@@ -125,6 +125,36 @@ where
...
+-- ORCA picks (A) - Producer and Consumer stay on the same slice and the
+-- bug does not manifest. As outer_rows grows, the per-row term of (A)
+-- scales linearly while the fixed overhead of (B) does not, so the
+-- curves cross and (B) becomes cheaper. 40000 rows in ss_t1 is 
+-- the smallest input we found that pushes outer_rows past that crossover 
+-- on typical segment counts; with fewer rows plan (A) wins on cost and the 
+-- unpaired-Consumer path is never entered, so the regression test would
+-- silently pass even without the fix.
+CREATE TABLE ss_t1 AS
+  SELECT i AS id FROM generate_series(1, 40000) i
```
andr-sokolov:
```suggestion
  SELECT generate_series(1, 40000) id
```

Alena0704: Added

src/backend/gporca/libgpopt/src/base/CUtils.cpp
```diff
@@ -901,6 +901,70 @@ CUtils::FHasCTEAnchor(CExpression *pexpr)
...
+	// check if every consumer's producer is in ProducerSet
+	for (ULONG ul = 0; ul < cteConsumers->Size(); ul++)
+	{
+		ULONG *pulCteId = (*cteConsumers)[ul];
+		if (cteProducerSet->Contains(pulCteId))
+		{
+			continue;
+		}
+		if (NULL != triggerOnlyForCteIds &&
+			!triggerOnlyForCteIds->Contains(pulCteId))
```
andr-sokolov: What about inverting this conditions and removing `continue`?

src/test/regress/sql/qp_orca_fallback.sql
```diff
@@ -126,3 +126,29 @@ INSERT INTO partition_key_dropped VALUES(3);
...
+	referenceid numeric NULL
+)
+DISTRIBUTED REPLICATED;
+CREATE TABLE tbl1 (
+	id bigserial NOT NULL,
+	iscalctrg varchar(15) NOT NULL,
+	iscalcdetail varchar(15) NULL
+)
+DISTRIBUTED REPLICATED;
+-- end_ignore
```
andr-sokolov: Why do we need `end_ignore` here?

src/backend/gporca/libgpopt/src/translate/CTranslatorExprToDXL.cpp
```diff
@@ -3427,7 +3441,7 @@ CTranslatorExprToDXL::PdxlnExistentialSubplan(
 	// translate inner child
 	CExpression *pexprInner = (*pexprCorrelatedNLJoin)[1];
 
-	CDXLNode *pdxlnInnerChild = CreateDXLNode(
+CDXLNode *pdxlnInnerChild = CreateDXLNode(
```
andr-sokolov: unrelated change

src/backend/gporca/libgpopt/src/translate/CTranslatorExprToDXL.cpp
```diff
@@ -3190,7 +3204,7 @@ CTranslatorExprToDXL::PdxlnQuantifiedSubplan(
 	CExpression *pexprInner = (*pexprCorrelatedNLJoin)[1];
 	CExpression *pexprScalar = (*pexprCorrelatedNLJoin)[2];
 
-	// translate inner child
+// translate inner child
```
andr-sokolov: unrelated change

src/backend/gporca/libgpopt/src/translate/CTranslatorExprToDXL.cpp
```diff
@@ -3492,7 +3506,11 @@ CTranslatorExprToDXL::BuildScalarSubplans(
 {
 	const ULONG size = pdrgpcrInner->Size();
 
-	CDXLNodeArray *pdrgpdxlnInner = GPOS_NEW(m_mp) CDXLNodeArray(m_mp);
+	// Fallback to Postgres optimizer if the SubPlan's inner expression contains a
+	// CTE Consumer whose Producer lives outside this subtree. Such a Consumer
+	// would become a cross-slice Shared Scan reader without a local Producer,
+	// which can hang the query or fail at execution time.
+CDXLNodeArray *pdrgpdxlnInner = GPOS_NEW(m_mp) CDXLNodeArray(m_mp);
```
andr-sokolov: why is the indentation changed?

Alena0704: it wasn't necessary - fixed

src/backend/gporca/libgpopt/src/translate/CTranslatorExprToDXL.cpp
Alena0704: it wasn't necessary - fixed

src/backend/gporca/libgpopt/src/base/CUtils.cpp
```diff
@@ -901,6 +901,178 @@ CUtils::FHasCTEAnchor(CExpression *pexpr)
...
+CUtils::FHasCrossSliceReplicatedCTEConsumer(CMemoryPool *mp, CExpression *pexpr)
+{
+	if (NULL == pexpr)
+	{
+		return false;
+	}
+
+	ULongPtrArray *prodIds = GPOS_NEW(mp) ULongPtrArray(mp);
+	ULongPtrArray *prodSlices = GPOS_NEW(mp) ULongPtrArray(mp);
+	ULongPtrArray *prodReplicated = GPOS_NEW(mp) ULongPtrArray(mp);
```
andr-sokolov: ULongPtrArray contains uint32_t elements, but we use them as bool. What about declaring and using
```
typedef CDynamicPtrArray<BOOL, CleanupDelete> BoolPtrArray;
``` 
?

Can we use array of bool instead of array of pointers to bool/uint32_t?

src/backend/gporca/libgpopt/src/base/CUtils.cpp
```diff
@@ -901,6 +901,178 @@ CUtils::FHasCTEAnchor(CExpression *pexpr)
...
+	COperator *pop = pexpr->Pop();
+
+	if (COperator::EopPhysicalCTEProducer == pop->Eopid())
+	{
+		prodIds->Append(GPOS_NEW(mp) ULONG(
+			CPhysicalCTEProducer::PopConvert(pop)->UlCTEId()));
+		prodSlices->Append(GPOS_NEW(mp) ULONG(curSlice));
+
+		BOOL replicated = false;
+		if (1 == pexpr->Arity())
```
andr-sokolov: Why is the distribution not replicated when `pexpr->Arity() != 1`?

src/backend/gporca/libgpopt/src/base/CUtils.cpp
```diff
@@ -901,6 +901,135 @@ CUtils::FHasCTEAnchor(CExpression *pexpr)
...
+			SCTEProducerInfo *info = (*prodInfos)[ip];
+			if (info->cteId != cid)
+			{
+				continue;
+			}
+			if (info->isReplicated && info->sliceId != cslice)
+			{
+				cross = true;
+			}
+			break;
```
andr-sokolov: can we simplify this fragment?
```suggestion
			if (info->cteId == cid)
			{
			    cross = info->isReplicated && info->sliceId != cslice;
			    break;
			}
```

src/backend/gporca/libgpopt/src/base/CUtils.cpp
```diff
@@ -901,6 +901,135 @@ CUtils::FHasCTEAnchor(CExpression *pexpr)
...
+		// Allocate a fresh slice id for each non-scalar child of a
+		// Motion; otherwise the child stays in the parent's slice.
+		ULONG childSlice = curSlice;
+		if (isMotion)
+		{
+			(*pNextSlice)++;
+			childSlice = *pNextSlice;
+		}
+
+		CollectCTESlices(mp, pexprChild, childSlice, pNextSlice, prodInfos,
```
andr-sokolov, line 982: In which case do childSlice and *pNextSlice have different values?

src/backend/gporca/libgpopt/src/base/CUtils.cpp
```diff
@@ -901,6 +901,131 @@ CUtils::FHasCTEAnchor(CExpression *pexpr)
...
+		CExpression *pexprChild = (*pexpr)[0];
+		CDrvdPropPlan *pdpplan =
+			CDrvdPropPlan::Pdpplan(pexprChild->PdpDerive());
+
+		SCTEProducerInfo *info = GPOS_NEW(mp) SCTEProducerInfo;
+		info->cteId = CPhysicalCTEProducer::PopConvert(pop)->UlCTEId();
+		info->sliceId = curSlice;
+		info->isReplicated =
+			FReplicatedLikeDistribution(pdpplan->Pds()->Edt());
+		prodInfos->Append(info);
```
andr-sokolov: Can we remove isReplicated from SCTEProducerInfo? Can we use the same structure to save information about producers and consumers?
```suggestion
		if (FReplicatedLikeDistribution(pdpplan->Pds()->Edt()))
			prodInfos->Append(info);
```

src/backend/gporca/libgpopt/src/base/CUtils.cpp
```diff
@@ -901,6 +901,131 @@ CUtils::FHasCTEAnchor(CExpression *pexpr)
...
+		for (ULONG ip = 0; ip < prodInfos->Size(); ip++)
+		{
+			SCTEProducerInfo *info = (*prodInfos)[ip];
+			if (info->cteId == cid)
+			{
+			    cross = info->isReplicated && info->sliceId != cslice;
+			    break;
+			}
+		}
+	}
```
andr-sokolov: I suggest to use goto, because it is more explicit way to say that we should break the both loops. Also we rid of redundant checks of the cross variable
```suggestion
	for (ULONG ic = 0; ic < consIds->Size(); ic++)
	{
		ULONG cid = *(*consIds)[ic];
		ULONG cslice = *(*consSlices)[ic];

		for (ULONG ip = 0; ip < prodInfos->Size(); ip++)
		{
			SCTEProducerInfo *info = (*prodInfos)[ip];
			if (info->cteId == cid && info->sliceId != cslice)
			{
			    cross = true;
			    goto lExit;
			}
		}
	}

lExit:
```

andr-sokolov (approved):

PR #376, opened 2026-04-20
[subquery] Preserve COUNT semantics in scalar subquery pull-up

### LINKED ISSUE #378: [subquery] COUNT-containing scalar subquery rewrite can return wrong results in fallback PostgreSQL optimizer path
### Greenplum version or build
open-gpdb `OPENGPDB_STABLE` (GP6/PG9.4-based line)

### Expected behavior
Correlated scalar aggregate subqueries should preserve empty-input semantics after pull-up rewrite.

If target expression contains `COUNT`, unmatched rows must evaluate with aggregate empty-input defaults (e.g. `COUNT -> 0`).

### Actual behavior
In fallback PostgreSQL optimizer rewrite path (`cdbsubselect.c`, `convert_EXPR_to_join`), COUNT-containing scalar-subquery expressions can return incorrect results on unmatched rows.

This is **not ORCA-specific**. It affects execution paths that fall back to PostgreSQL planner.

### Step to reproduce the behavior
```sql
CREATE TABLE t_outer(a int, b int) DISTRIBUTED BY (a);
CREATE TABLE t_inner(b int) DISTRIBUTED BY (b);
INSERT INTO t_outer VALUES (0,1),(0,2);
INSERT INTO t_inner VALUES (1);

set optimizer=off;
SELECT a,b
FROM t_outer o
WHERE (o.a + 1) > (
  SELECT count(*) + 1
  FROM t_inner i
  WHERE i.b = o.b
)
ORDER BY b;
```

### Additional impact scope
Besides `optimizer=off`, the same risk exists for other query shapes that trigger ORCA fallback to PostgreSQL planner (for example unsupported features/GUC combinations where planning switches to Postgres optimizer).

So the issue scope is: **fallback PostgreSQL planner rewrite path**, not only manual ORCA-off sessions.

Excaliiiibur:
Fixes #378

## Summary
This is not an ORCA issue; it occurs in the fallback PostgreSQL optimizer rewrite path.

Fix incorrect semantics in correlated scalar aggregate subquery pull-up when target expression contains `COUNT`.

This risk applies not only to manual `optimizer=off`, but also to other cases where planning falls back from ORCA to PostgreSQL optimizer.

## What Changed
In `cdbsubselect.c` (`convert_EXPR_to_join`), for COUNT-containing scalar aggregate expressions:
- use `LEFT JOIN` in rewrite
- replace pulled-up expression with `COALESCE(agg_expr, default_expr)`
- derive `default_expr` from empty-input aggregate defaults:
  - `COUNT(...) -> 0`
  - non-COUNT aggregates -> `NULL` (aggregate result type)

This preserves no-match semantics for both COUNT-only and mixed COUNT+other aggregate expressions.

## Tests Updated
Added/updated `subselect_gp` regression cases for:
- `count(*) + 1`
- `(count(*) + 1)::bigint`
- `case when count(*) > 0 then count(*) + 1 else 1 end`
- `abs(count(*) - 1)`
- `count(*) + coalesce(sum(...), 0)`
- `coalesce(count(*) + sum(...), -1)`

## Files
- `src/backend/cdb/cdbsubselect.c`
- `src/test/regress/sql/subselect_gp.sql`
- `src/test/regress/expected/subselect_gp.out`
- `src/test/regress/expected/subselect_gp_optimizer.out`

### DIFF
--- a/src/backend/cdb/cdbsubselect.c
+++ b/src/backend/cdb/cdbsubselect.c
@@ -17,6 +17,7 @@
 #include "access/htup_details.h"
 #include "access/skey.h"
 #include "catalog/pg_operator.h"
+#include "catalog/pg_proc.h"
 #include "catalog/pg_type.h"
 #include "nodes/makefuncs.h"
 #include "optimizer/clauses.h"
@@ -43,6 +44,10 @@ static JoinExpr *make_join_expr(Node *larg, int r_rtindex, int join_type);
 static Node *make_lasj_quals(PlannerInfo *root, SubLink *sublink, int subquery_indx);
 
 static Node *add_null_match_clause(Node *clause);
+static bool contains_count_agg_expr(Node *expr);
+static Expr *build_count_coalesce_expr(Var *aggVar, Expr *defaultExpr);
+static Expr *build_count_default_expr(Node *expr);
+static Node *replace_agg_with_empty_default_mutator(Node *node, void *context);
 
 typedef struct NonNullableVarsContext
 {
@@ -574,6 +579,7 @@ convert_EXPR_to_join(PlannerInfo *root, OpExpr *opexp)
 		 * targetlist.
 		 */
 		TargetEntry *origSubqueryTLE = (TargetEntry *) list_nth(subselect->targetList, 0);
+		bool		has_count_expr = contains_count_agg_expr((Node *) origSubqueryTLE->expr);
 
 		List	   *subselectTargetList = (List *) copyObject(ctx1.targetList);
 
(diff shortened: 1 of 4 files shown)


Alena0704:
As I see, the case like this (SELECT COUNT(*)+1 FROM ...) is still incorrect. You should add the check that the type of expression here can be OpExpr in the is_plain_count_agg_expr function.

Is this case is reproducible for orca? we should mention it in the commit message if it isn't and only for fallback postgres optimizer (just in case).

Excaliiiibur:
Thanks for the feedback.

I’ve updated the fix to cover non-plain COUNT expressions as well, not just plain COUNT(*). This now includes cases like `count(*) + 1`, casts, `CASE`, `abs(...)`, and mixed expressions such as `count(*) + coalesce(sum(...), 0)`.

Implementation-wise, for COUNT-containing scalar aggregate expressions in the fallback PostgreSQL rewrite path, we preserve no-match semantics via `LEFT JOIN + COALESCE(agg_expr, default_expr)`, where defaults are derived as `COUNT -> 0` and non-COUNT aggregates -> `NULL` of aggregate result type.

I also added regression coverage for these cases, including fallback-path validation with `optimizer=on` + ORCA fallback. So this is not ORCA-native behavior; it is specific to the fallback PostgreSQL optimizer path.

Could you please take another look?

Alena0704:
> Thanks for the feedback.
> 
> I’ve updated the fix to cover non-plain COUNT expressions as well, not just plain COUNT(*). This now includes cases like `count(*) + 1`, casts, `CASE`, `abs(...)`, and mixed expressions such as `count(*) + coalesce(sum(...), 0)`.
> 
> Implementation-wise, for COUNT-containing scalar aggregate expressions in the fallback PostgreSQL rewrite path, we preserve no-match semantics via `LEFT JOIN + COALESCE(agg_expr, default_expr)`, where defaults are derived as `COUNT -> 0` and non-COUNT aggregates -> `NULL` of aggregate result type.
> 
> I also added regression coverage for these cases, including fallback-path validation with `optimizer=on` + ORCA fallback. So this is not ORCA-native behavior; it is specific to the fallback PostgreSQL optimizer path.
> 
> Could you please take another look?

Yes, sure)

Alena0704:
Hi! I am confused, what commit I should see. This commit https://github.com/Excaliiiibur/gpdb/pull/1/changes/df87bf0f44670185aac813340a5c6de7b3e92c46, right?
I have noticed the notification that your branch was outdated and pushed the button yo update it. I noticed there is diff in the regression tests - please, take a look at them and fix them. You need to open the label "check and display regression diffs" - I think they was caused by your patch - you need to fix it anyway.

Alena0704:
Because of lack of access to your branches and it is impossible to change the branch in PR, I opened the separate PR https://github.com/open-gpdb/gpdb/pull/397.

PR #388, opened 2026-06-11
Fix: error on creating try_convert extension

roaldm153:
Non-superusers can't create `try_convert` extension, receiving a permission denied error.
The fix is to add `trusted = true` to try_convert.control.
`trusted = true` allows installing the extension to any users with CREATE privileges on current database. 
Extension is safe.
- [ ] Document changes

### DIFF
--- a/contrib/try_convert/try_convert.control
+++ b/contrib/try_convert/try_convert.control
@@ -1,5 +1,6 @@
 # try_convert extension
 comment = 'function for type cast'
-default_version = '1.0'
+default_version = '1.1'
 module_pathname = '$libdir/try_convert'
 relocatable = true
+trusted = true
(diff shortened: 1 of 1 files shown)


contrib/try_convert/try_convert.control
```diff
@@ -1,5 +1,6 @@
 # try_convert extension
 comment = 'function for type cast'
-default_version = '1.0'
+default_version = '1.1'
```
Smyatkin-Maxim: No need for this as we don't change .sql file.

Smyatkin-Maxim (changes requested):
1.1 should be removed, otherwise it's ok

Smyatkin-Maxim (approved):

PR #389, opened 2026-06-13
orca: fall back for replicated CTE consumed in correlated scalar subqueries

Alena0704:
orca: fall back for replicated CTE consumed in correlated scalar subqueries

Root cause is a blind spot in the #375 slice walker: CollectCTESlices delimits slices only at Motion nodes. A CTE over a DISTRIBUTED REPLICATED table referenced from correlated scalar subqueries is decorrelated by ORCA into CPhysicalCorrelated*NLJoin whose inner side becomes an executor SubPlan running in its own slice -- but there is no Motion at that boundary, so the walker placed the Consumer on the same slice as the Producer. The cross-slice check (prod->sliceId != cons->sliceId) never fired, no fallback happened, and the ShareInputScan writer hung forever in shareinput_writer_waitdone() waiting for DONE acks from reader slices that never run.

Teach the walker that the inner (subquery) side of a correlated NL join is a slice boundary too, mirroring the Motion rule. The replicated Consumer in the SubPlan then gets a distinct slice id, the existing check fires, and ORCA falls back to the Postgres optimizer.

### DIFF
--- a/src/backend/gporca/libgpopt/src/base/CUtils.cpp
+++ b/src/backend/gporca/libgpopt/src/base/CUtils.cpp
@@ -922,10 +922,30 @@ struct SCTEInfo
 
 typedef CDynamicPtrArray<SCTEInfo, CleanupDelete<SCTEInfo> > CTEInfoArray;
 
+// True if the operator is a correlated NL join. Its inner side becomes
+// an executor SubPlan that runs in its own slice, so a CTE Consumer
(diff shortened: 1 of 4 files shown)


Alena0704:
I have found a commit in greengage that can fix this problem. I am going to fix this there: https://github.com/open-gpdb/gpdb/pull/396

PR #391, opened 2026-06-29
Add OPENGPDB compile-time identification macro

Alena0704:
Add OPENGPDB compile-time identification macro

Define OPENGPDB in pg_config_manual.h so extensions can tell open-gpdb apart from other Greenplum forks (e.g. GreengageDB) at compile time.

Extensions need this because internal APIs differ between forks, so the same source must pick a different call depending on the fork. For example utils/sampling.h: open-gpdb has sampler_random_fract(void), GreengageDB has sampler_random_fract(SamplerRandomState). With this macro an extension can guard such calls via #ifdef OPENGPDB.

### DIFF
--- a/src/include/pg_config_manual.h
+++ b/src/include/pg_config_manual.h
@@ -311,3 +311,10 @@
 /* #define HEAPDEBUGALL */
 /* #define ACLDEBUG */
 /* #define RTDEBUG */
+
+/*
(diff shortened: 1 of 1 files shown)


andr-sokolov (approved):

leborchuk (approved):
LGTM

PR #393, opened 2026-07-02
Fix flaky ci tests

Alena0704:
Fix flaky parallel_retrieve_cursor/status_check test

Test9 relied on the segment executors scanning t1 slower than DECLARE
returns, so on fast runners the sqrt(-1) error raced DECLARE and surfaced
synchronously. Add a FAULT_INJECTOR hook parallel_retrieve_cursor_before_exec
in standard_ExecutorRun (after the endpoint is created and the coordinator
is notified, before ExecutePlan) and have the test pause the segment
executors there, so the error is always raised after DECLARE and detected
only via the timeout mechanism.

Fix flaky ao_fsync_panic test (failover-window race)
The test injected checkpoint_after_redo_calculated on content 0 right after
pg_ctl stop immediate + FTS probe, hitting the promoting mirror while it still
refuses connections ("connection to dbid ... failed"). Arm it on the mirror
before the stop instead (it only fires in CreateCheckPoint(), so promotion
still triggers it) and wait for the new primary via "0U: SELECT 1;".

### DIFF
--- a/src/backend/executor/execMain.c
+++ b/src/backend/executor/execMain.c
@@ -1152,6 +1152,19 @@ standard_ExecutorRun(QueryDesc *queryDesc,
 				 */
 				EndpointNotifyQD(ENDPOINT_READY_ACK_MSG);
 
+#ifdef FAULT_INJECTOR
+				/*
+				 * Testing hook: DECLARE has already returned (endpoint created,
+				 * QD notified) but the plan hasn't run yet. A test pauses here so
+				 * an execution error surfaces only after DECLARE and is caught by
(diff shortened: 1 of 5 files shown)


Alena0704:
It is likely CI runner CPU contention, not a code issue. Jobs run on ubuntu-latest (4 vCPU) but the demo cluster is 7 Postgres processes (3 primaries + 3 mirrors + coordinator). Under that load, FTS promotion and crash recovery get slow and racy - hence the transient connection to dbid N ... failed during promotion. A different recovery test flakes each run (status_check → ao_fsync_panic → pg_rewind), which points to contention too.

PR #395, opened 2026-07-02
Fix CI: add S3 credentials for OTM buckets in yproxy.conf

Alena0704:
Fix CI: add S3 credentials for OTM buckets in yproxy.conf

The OTM buckets gpyezzey2/gpyezzey3 had no credentials, so yproxy rejected offload into the tab1/tab2 tablespaces. Add a credential_map with the access keys for all buckets to fix it.


leborchuk (approved):
LGTM

PR #396, opened 2026-07-02
Fix cross-slice interaction detection for subplans

Alena0704:
Fix cross-slice interaction detection for subplans
    
GPDB has Shared Scan node to share the output of a subplan. Each Shared Scan
node can act like producer or consumer. Each one may be a part of different
slice. One of essential parts of Shared Scan processing is cross-slice
interaction detection. It helps to find the difference in producer's and
consumer's slices and mark underlying Material or Sort nodes as cross-slice.
Cross-slice producer reads all the tuples from child plan before the consumer
Shared Scan can read it.

Before this patch, cross-slice interaction detection of underlying subplan nodes
was broken. We walked through subplans separately from main plan. This, in case
subplan has no motion, caused `shareinput_mutator_xslice_2()` to think we don't
need to mark node as cross-slice - motion stack (`motStack`) used for
cross-slice interaction detection contained only default motion id (0) which
equals to producer's slice id.

Here is the simplified part of new regression test. It shows Shared Scan
consumer under Subplan, which has different (compared to producer upper) slice
id.
```
Sequence
 ->  Shared Scan (share slice:id 0:0)
   ->  Seq Scan on t1 t1_1
 ->  Gather Motion 3:1  (slice3; segments: 3)
   ->  Redistribute Motion 1:3  (slice1)
     ->  Function Scan on generate_series
       SubPlan 1  (slice1)
         ->  Shared Scan (share slice:id 1:0)
```
Before this patch, we didn't have upper motions in `motStack`. From now, we dive
into subplans as parts of main plan according to the plan tree hierarchy.
Cross-slice detector understands that our slice(1) is different from producer's
slice(0) and marks node as cross-slice.

Existing logic and comments around rtables left almost unchanged.
Existing tricky logic around joins execution order changed to another tricky
logic which is compatible with `plan_tree_walker()`.
`gporca_optimizer.out` test output changed as we now start
`shareinput_mutator_dag_to_tree()` from the main plan. Before, we started from
subplans and global share input context contained only 3 producers at the time
`set_plan_share_id()` called. Now, at the same place producers count is equal to
4, so RangeTblEntry's name changed.

There was a bug when duplicate subplans were added without new corresponded
subroots. The old shareinput_walker() implementation not called for all
subplans, because forboth() loop stopped early. New shareinput_walker()
implementation not found appropriate subplan's subroot (and used main root as
default). From now, we fix old mistake and create new subroot for each new
subplan.

Author: Alexey Gordeev <lexa-first@ya.ru>
cherry-pick from a39683f1b74 #342 (https://github.com/GreengageDB/greengage)

### DIFF
--- a/src/backend/cdb/cdbmutate.c
+++ b/src/backend/cdb/cdbmutate.c
@@ -2947,20 +2947,21 @@ fixup_subplan_walker(Node *node, SubPlanWalkerContext *context)
 			 * there is more than one subplan node which refer to the same
 			 * plan_id. In this case create a duplicate subplan, append it to
 			 * the glob->subplans and update the plan_id of the subplan to
-			 * refer to the new copy of the subplan node. Note since subroot
-			 * is not used there is no need of new subroot, and initplans are
+			 * refer to the new copy of the subplan node. glob->subroots must
+			 * stay aligned with glob->subplans (same length, indexed by
+			 * plan_id); the duplicate's subroot is only ever read (e.g. by
+			 * apply_shareinput_xslice), never mutated through this slot, so we
+			 * reuse the original subroot instead of copying it. Initplans are
 			 * not needed to be duplicated.
 			 */
 			PlannerInfo *root = (PlannerInfo *) context->base.node;
 			Plan	    *dupsubplan = (Plan *) copyObject(planner_subplan_get_plan(root, subplan));
+			PlannerInfo *subroot = planner_subplan_get_root(root, subplan);
 			int			 newplan_id = list_length(root->glob->subplans) + 1;
-			PlannerInfo *dupsubroot = makeNode(PlannerInfo);
-
-			memcpy(dupsubroot, planner_subplan_get_root(root, subplan), sizeof(PlannerInfo));
-			root->glob->subroots = lappend(root->glob->subroots, dupsubroot);
 
 			subplan->plan_id = newplan_id;
 			root->glob->subplans = lappend(root->glob->subplans, dupsubplan);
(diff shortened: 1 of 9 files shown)


src/backend/cdb/cdbmutate.c
```diff
@@ -3001,6 +2954,10 @@ fixup_subplan_walker(Node *node, SubPlanWalkerContext *context)
 			PlannerInfo *root = (PlannerInfo *) context->base.node;
 			Plan	    *dupsubplan = (Plan *) copyObject(planner_subplan_get_plan(root, subplan));
 			int			 newplan_id = list_length(root->glob->subplans) + 1;
+			PlannerInfo *dupsubroot = makeNode(PlannerInfo);
+
+			memcpy(dupsubroot, planner_subplan_get_root(root, subplan), sizeof(PlannerInfo));
```
andr-sokolov: I suggest to use `=` instead of `memcpy` to let compiler check types and avoid errors with sizes, and replace `makeNode` with `palloc`, because all the content of `dupsubroot` is changed in the next line.
```
			PlannerInfo *dupsubroot = palloc(sizeof(*dupsubroot));

			*dupsubroot = *(planner_subplan_get_root(root, subplan));
```

Do we really need memory allocation here?

Alena0704: Good catch - we don't need the allocation at all. The duplicate's subroot is only ever read (e.g. by apply_shareinput_xslice), never mutated through its glob->subroots slot, and makeNode+memcpy was only a shallow copy anyway. Comment updated accordingly. Done in a follow-up commit.

andr-sokolov (approved):

PR #400, opened 2026-07-07
Don't use Orca to plan SELECTs with an empty range table

Alena0704:
Skip Orca for a SELECT with an empty range table, no sublinks and no parent
utility statement.  This covers FROM-less SQL function bodies too. The rule also 
applies to the bodies of SQL functions.  That is why some Orca fallback messages 
disappear even for queries that do have a FROM clause: the message came from 
planning the function body, not the query execution itself.

For some of these queries - the ones with parameters or with SIRV functions - 
Orca does not produce a plan at all and falls back, so the attempt is wasted twice over.  
Skipping it makes such queries about 16x faster.

Co-Authored-By: andr-sokolov [sokolov.andrey.yurevich@gmail.com](mailto:sokolov.andrey.yurevich@gmail.com)

### DIFF
--- a/gpcontrib/pgaudit/expected/pgaudit.out
+++ b/gpcontrib/pgaudit/expected/pgaudit.out
@@ -795,10 +795,9 @@ NOTICE:  AUDIT: SESSION,31,2,READ,SELECT,,,explain select 1;,<none>
 NOTICE:  AUDIT: SESSION,31,3,MISC,EXPLAIN,,,explain select 1;,<none>
                 QUERY PLAN                
 ------------------------------------------
- Result  (cost=0.00..0.00 rows=1 width=4)
-   ->  Result  (cost=0.00..0.00 rows=1 width=1)
- Optimizer: Pivotal Optimizer (GPORCA)
-(3 rows)
+ Result  (cost=0.00..0.01 rows=1 width=0)
+ Optimizer: Postgres query optimizer
+(2 rows)
 
 --
 -- Test that looks inside of do blocks log
@@ -880,12 +879,11 @@ BEGIN
 	EXECUTE 'DROP table ' || table_name;
 END $$;",<none>
 NOTICE:  AUDIT: SESSION,36,2,AST_SEL,SELECT,,,{QUERY...},<none>
-NOTICE:  AUDIT: SESSION,36,3,READ,SELECT,,,SELECT 'do_table',<none>
-NOTICE:  AUDIT: SESSION,36,4,AST_SEL,SELECT,,,{QUERY...},<none>
+NOTICE:  AUDIT: SESSION,36,3,AST_SEL,SELECT,,,{QUERY...},<none>
 NOTICE:  Table doesn't have 'DISTRIBUTED BY' clause -- Using column named 'weird name' as the Greenplum Database data distribution key for this table.
-NOTICE:  AUDIT: SESSION,36,5,DDL,CREATE TABLE,TABLE,public.do_table,"CREATE TABLE do_table (""weird name"" INT)",<none>
-NOTICE:  AUDIT: SESSION,36,6,AST_SEL,SELECT,,,{QUERY...},<none>
-NOTICE:  AUDIT: SESSION,36,7,DDL,DROP TABLE,TABLE,public.do_table,DROP table do_table,<none>
+NOTICE:  AUDIT: SESSION,36,4,DDL,CREATE TABLE,TABLE,public.do_table,"CREATE TABLE do_table (""weird name"" INT)",<none>
+NOTICE:  AUDIT: SESSION,36,5,AST_SEL,SELECT,,,{QUERY...},<none>
+NOTICE:  AUDIT: SESSION,36,6,DDL,DROP TABLE,TABLE,public.do_table,DROP table do_table,<none>
 --
 -- Generate an error and make sure the stack gets cleared
 DO $$
@@ -1087,11 +1085,10 @@ NOTICE:  AUDIT: SESSION,60,1,AST_SEL,SELECT,,,{QUERY...}{f XXXX public.test {}},
 NOTICE:  AUDIT: SESSION,60,2,READ,SELECT,,,SELECT test();,<none>
 NOTICE:  AUDIT: SESSION,60,3,FUNCTION,EXECUTE,FUNCTION,public.test,SELECT test();,<none>
 NOTICE:  AUDIT: SESSION,60,4,AST_SEL,SELECT,,,{QUERY...},<none>
-NOTICE:  AUDIT: SESSION,60,5,READ,SELECT,,,SELECT 'cur1'::pg_catalog.refcursor,<none>
-NOTICE:  AUDIT: SESSION,60,6,AST_SEL,SELECT,,,"{QUERY...}{r XXXX public.hoge {id}}",<none>
-NOTICE:  AUDIT: SESSION,60,6,AST_SEL,SELECT,TABLE,public.hoge,"{QUERY...}{r XXXX public.hoge {id}}",<none>
-NOTICE:  AUDIT: SESSION,60,7,READ,SELECT,TABLE,public.hoge,select * from hoge,<none>
-NOTICE:  AUDIT: SESSION,60,8,AST_SEL,SELECT,,,{QUERY...},<none>
+NOTICE:  AUDIT: SESSION,60,5,AST_SEL,SELECT,,,"{QUERY...}{r XXXX public.hoge {id}}",<none>
+NOTICE:  AUDIT: SESSION,60,5,AST_SEL,SELECT,TABLE,public.hoge,"{QUERY...}{r XXXX public.hoge {id}}",<none>
+NOTICE:  AUDIT: SESSION,60,6,READ,SELECT,TABLE,public.hoge,select * from hoge,<none>
+NOTICE:  AUDIT: SESSION,60,7,AST_SEL,SELECT,,,{QUERY...},<none>
  test 
 ------
      
@@ -1232,10 +1229,9 @@ NOTICE:  AUDIT: SESSION,72,1,AST_SEL,SELECT,,,"{QUERY...}{f XXXX public.get_out_
 NOTICE:  AUDIT: SESSION,72,2,READ,SELECT,,,SELECT * FROM get_out_args(3);,<none>
 NOTICE:  AUDIT: SESSION,72,3,FUNCTION,EXECUTE,FUNCTION,public.get_out_args,SELECT * FROM get_out_args(3);,<none>
 NOTICE:  AUDIT: SESSION,72,4,AST_SEL,SELECT,,,{QUERY...},<none>
-NOTICE:  AUDIT: SESSION,72,5,READ,SELECT,,,SELECT 1,<none>
+NOTICE:  AUDIT: SESSION,72,5,AST_SEL,SELECT,,,{QUERY...},<none>
 NOTICE:  AUDIT: SESSION,72,6,AST_SEL,SELECT,,,{QUERY...},<none>
 NOTICE:  AUDIT: SESSION,72,7,AST_SEL,SELECT,,,{QUERY...},<none>
-NOTICE:  AUDIT: SESSION,72,8,AST_SEL,SELECT,,,{QUERY...},<none>
  col_o1 | col_o2 
 --------+--------
       2 |      3
@@ -1273,10 +1269,9 @@ NOTICE:  AUDIT: SESSION,74,1,AST_SEL,SELECT,,,"{QUERY...}{f XXXX public.get_tabl
 NOTICE:  AUDIT: SESSION,74,2,READ,SELECT,,,SELECT * FROM get_table(2);,<none>
 NOTICE:  AUDIT: SESSION,74,3,FUNCTION,EXECUTE,FUNCTION,public.get_table,SELECT * FROM get_table(2);,<none>
 NOTICE:  AUDIT: SESSION,74,4,AST_SEL,SELECT,,,{QUERY...},<none>
-NOTICE:  AUDIT: SESSION,74,5,READ,SELECT,,,SELECT 1,<none>
+NOTICE:  AUDIT: SESSION,74,5,AST_SEL,SELECT,,,{QUERY...},<none>
 NOTICE:  AUDIT: SESSION,74,6,AST_SEL,SELECT,,,{QUERY...},<none>
 NOTICE:  AUDIT: SESSION,74,7,AST_SEL,SELECT,,,{QUERY...},<none>
-NOTICE:  AUDIT: SESSION,74,8,AST_SEL,SELECT,,,{QUERY...},<none>
  col_t1 | col_t2 
 --------+--------
       2 |      3
--- a/src/backend/optimizer/plan/planner.c
+++ b/src/backend/optimizer/plan/planner.c
@@ -157,6 +157,7 @@ static Plan *pushdown_preliminary_limit(Plan *plan, Node *limitCount, int64 coun
 
 static Plan *getAnySubplan(Plan *node);
 static bool isSimplyUpdatableQuery(Query *query);
+static bool isQueryForOrca(Query *parse);
 
 
 /*****************************************************************************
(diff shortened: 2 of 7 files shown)


src/backend/optimizer/plan/planner.c
```diff
@@ -5547,50 +5554,50 @@ static Bitmapset* canonicalize_colref_list(Node * node)
 	ListCell *lc;
 	SortGroupClause *gc;
 	Bitmapset* gs = NULL;
-
```
andr-sokolov, line 5550: The PR contains a lot of unrelated changes. Please remove them

src/backend/optimizer/plan/planner.c
```diff
@@ -229,11 +230,17 @@ standard_planner(Query *parse, int cursorOptions, ParamListInfo boundParams)
...
+	 * which have no range table to distribute.  ORCA brings no benefit for
+	 * such queries and merely adds optimization overhead before falling back
+	 * to the Postgres planner anyway.
 	 */
 	if (optimizer &&
 		GP_ROLE_DISPATCH == Gp_role &&
 		IS_QUERY_DISPATCHER() &&
-		(cursorOptions & CURSOR_OPT_PARALLEL_RETRIEVE) == 0)
+		(cursorOptions & CURSOR_OPT_PARALLEL_RETRIEVE) == 0 &&
+		!queryIsConstantSelect(parse))
```
andr-sokolov: I hope we will add new rules about skipping Orca later. So I suggest to rename the function to something like `isQueryForOrca` and we will call only this function here.

src/backend/optimizer/plan/planner.c
```diff
@@ -5861,3 +5868,16 @@ isSimplyUpdatableQuery(Query *query)
...
+ * queryIsConstantSelect
+ *  determine whether a query is a trivial constant SELECT.
+ */
+static bool
+queryIsConstantSelect(Query *parse)
+{
+	return parse->commandType == CMD_SELECT &&
+		parse->rtable == NIL &&
+		!parse->hasSubLinks &&
+		parse->parentStmtType == PARENTSTMTTYPE_NONE;
```
andr-sokolov: I think we should somehow change this condition to don't affect this test
```
 select aggfns(a,b,c)
   from (values (1,3,'foo'),(0,null,null),(2,2,'bar'),(3,1,'baz')) v(a,b,c);
-INFO:  GPORCA failed to produce a plan, falling back to planner
-DETAIL:  Feature not supported: ROW EXPRESSION
-CONTEXT:  SQL function "aggfns_trans" during startup
```

src/backend/optimizer/plan/planner.c
Alena0704: Agree. Removed

src/backend/optimizer/plan/planner.c
Alena0704: Agree. Renamed

src/backend/optimizer/plan/planner.c
Alena0704: Fixed by adding a !query_has_params() check

Alena0704:
Just in case, I'll write about changes in out files of regression tests. All changes stem from one cause: trivial constant SELECTs with no range table (SELECT 1, SELECT pg_column_size(...), SELECT pg_lock_status()) now bypass ORCA and are planned by the Postgres planner. 

- bfv_catalog_optimizer.out. In EXPLAIN, the two nested Result nodes and Pivotal Optimizer (GPORCA) are replaced by a single Result and Optimizer: Postgres query optimizer. Queries are simply planned by Postgres now, the plan is simpler (one node instead of two), and the result is identical.
- qp_gist_indexes2_optimizer.source. The lines DETAIL: Feature not supported: SIRV functions and the second GPORCA failed to produce a plan are gone. Previously ORCA tried to plan the constant EXPLAIN sub-query and failed twice (first on SIRV functions, then on required properties), producing a doubled fallback message. Now that trivial query never goes to ORCA, so only the single honest fallback from the real query remains.
- pgaudit.out. This is the subtlest one. The READ,SELECT,,,SELECT 1 line disappeared. The READ record is emitted by the ExecutorStart hook (commandText = sourceText), while AST_SEL is emitted by the planner hook at plan time. For SELECT 1 INTO test inside the DO block: ORCA produced a plan with two Result nodes, which PL/pgSQL does not recognize as a "simple expression", so the query ran through the executor → ExecutorStart fired → a READ line appeared. The Postgres planner produces a single Result, which PL/pgSQL recognizes as a simple expression and evaluates directly, bypassing the executor → ExecutorStart never fires → no READ line. The AST_SEL line stays because planning (and thus the planner hook) still happens in both cases.

src/test/regress/expected/bfv_catalog_optimizer.out
```diff
@@ -257,22 +257,18 @@ reset optimizer_enable_indexscan;
...
- Settings:  optimizer=on
- Optimizer status: Pivotal Optimizer (GPORCA) version 2.23.0
-(4 rows)
+                QUERY PLAN                
+------------------------------------------
+ Result  (cost=0.00..0.01 rows=1 width=0)
+ Optimizer: Postgres query optimizer
+(2 rows)
 
 explain select pg_lock_status();
```
andr-sokolov, line 266: This query is not a trivial constant query, but Orca doesn't plan it now. Fix please the PR description

Alena0704: Agree, fixed

gpcontrib/pgaudit/expected/pgaudit.out
```diff
@@ -880,12 +879,11 @@ BEGIN
 	EXECUTE 'DROP table ' || table_name;
 END $$;",<none>
 NOTICE:  AUDIT: SESSION,36,2,AST_SEL,SELECT,,,{QUERY...},<none>
-NOTICE:  AUDIT: SESSION,36,3,READ,SELECT,,,SELECT 'do_table',<none>
-NOTICE:  AUDIT: SESSION,36,4,AST_SEL,SELECT,,,{QUERY...},<none>
+NOTICE:  AUDIT: SESSION,36,3,AST_SEL,SELECT,,,{QUERY...},<none>
```
andr-sokolov, line 882: Why is pgaudit test affected?

Alena0704:
For FROM-less SELECT queries ORCA has nothing to work with: there is no range table, nothing to distribute across segments, nothing to reorder. Its plan either matches the Postgres planner's but adds a redundant wrapping `Result` node and a worse cardinality estimate for set-returning functions, or it is not produced at all - ORCA fails on SIRV functions and falls back to the Postgres planner anyway. So bypassing ORCA here is not a "faster but worse plan" trade-off: the plan is strictly no worse, and better in node count and estimates, while we additionally save the entire optimization attempt - a FROM-less SELECT runs about 16x faster (0.33 ms -> 0.02 ms per query in a release build).

| Query | STABLE (ORCA) | Patched (PG planner) | Speedup |
|---|---|---|---|
| `SELECT 1` | 3,048 TPS · 0.328 ms | 51,563 TPS · 0.019 ms | **16.9x** |
| `SELECT 42, 'abc'::text, 1+2*3` | 2,492 TPS · 0.401 ms | 40,079 TPS · 0.025 ms | **16.1x** |
| `SELECT pg_column_size('...')` | 2,859 TPS · 0.350 ms | 45,546 TPS · 0.022 ms | **15.9x** |
| `SELECT count(*) FROM bench_t WHERE a > 10` (control) | 858 TPS | 849 TPS | 0.99x — unchanged |

gpcontrib/pgaudit/expected/pgaudit.out
Alena0704: ORCA plans a FROM-less SELECT as `Result -> Result`, and PL/pgSQL's "simple expression" fast path requires a single `Result` node. So inside PL/pgSQL these expressions were executed through SPI and the full executor, which triggered `ExecCheckRTPerms()` and hence pgaudit's `READ,SELECT` record. With the Postgres planner the plan is a single `Result`, the expression is evaluated inline — no executor, no record. Those records had no object at all (`READ,SELECT,,,SELECT 'do_table',<none>`), so nothing that actually reads a relation stopped being audited.

src/backend/optimizer/plan/planner.c
```diff
@@ -5861,3 +5867,53 @@ isSimplyUpdatableQuery(Query *query)
...
+}
+
+/*
+ * isQueryForOrca
+ *		Should this query be handed to the ORCA optimizer at all?
+ *
+ * ORCA brings no benefit for trivial constant queries such as "SELECT 42":
+ * they have no range table to distribute across segments, so the Postgres
+ * planner handles them directly, without the overhead of an ORCA attempt that
+ * would only fall back to the Postgres planner anyway.
```
andr-sokolov: Orca can plan "SELECT 42" without fallback

src/backend/optimizer/plan/planner.c
```diff
@@ -5861,3 +5867,53 @@ isSimplyUpdatableQuery(Query *query)
...
+/*
+ * query_has_params
+ *		Does the query tree reference any parameters?
+ */
+static bool
+query_has_params_walker(Node *node, void *context)
+{
+	if (node == NULL)
+		return false;
+	if (IsA(node, Param))
```
andr-sokolov: Why is this query for Orca?
```
PREPARE sel(int) as select $1;
execute sel(2);
```

src/backend/optimizer/plan/planner.c
```diff
@@ -5861,3 +5867,53 @@ isSimplyUpdatableQuery(Query *query)
...
+		return true;
+	if (IsA(node, Query))
```
andr-sokolov: rtable is NIL. Why is it possible that node is a Query?

src/backend/optimizer/plan/planner.c
```diff
@@ -5861,3 +5867,53 @@ isSimplyUpdatableQuery(Query *query)
...
+ * still handed to ORCA, preserving its existing handling of them.
+ *
+ * Additional rules for bypassing ORCA can be added here.
+ */
+static bool
+isQueryForOrca(Query *parse)
+{
+	if (parse->commandType == CMD_SELECT &&
+		parse->rtable == NIL &&
+		!parse->hasSubLinks &&
```
andr-sokolov, line 5890: Why does SubLink require Orca?
```
select exists(select 1);
```

src/backend/optimizer/plan/planner.c
Alena0704: Agree, fixed.

src/backend/optimizer/plan/planner.c
```diff
@@ -5861,3 +5867,53 @@ isSimplyUpdatableQuery(Query *query)
...
+ * Queries carrying parameters (for example the bodies of SQL/PL functions) are
+ * still handed to ORCA, preserving its existing handling of them.
+ *
+ * Additional rules for bypassing ORCA can be added here.
+ */
+static bool
+isQueryForOrca(Query *parse)
+{
+	if (parse->commandType == CMD_SELECT &&
+		parse->rtable == NIL &&
```
andr-sokolov, line 5889: What about condition that Query contains RangeTblEntry with type RTE_RELATION?

src/backend/optimizer/plan/planner.c
Alena0704: Since params change nothing here: a FROM-less query has nothing to distribute regardless, and by default Orca doesn't plan queries with params anyway (optimizer_enable_query_parameter is off), I dropped the check and updated the expected output.

src/backend/optimizer/plan/planner.c
Alena0704: The Query branch is not unreachable. query_tree_walker() with flags = 0 also walks cteList, and expression_tree_walker() passes cte->ctequery, consequently a Query and to the walker (see nodeFuncs.c, case T_CommonTableExpr). rtable can be NIL while cteList is not, when the CTE is never referenced and so gets no RTE: WITH c AS (SELECT 1) SELECT 42;. Without the IsA(node, Query) it would fall through to the default and throws an error "unrecognized node type".

src/backend/optimizer/plan/planner.c
Alena0704: An empty rtable is not enough when the query has sublinks: select exists(select 1 from t); also has rtable == NIL at the top level, but the subquery still scans a distributed table. That is why a query with sublinks is still handed to ORCA. select exists(select 1) is a false positive of that rule — we could recurse into the sublinks and require an empty rtable there too, but I don't think it is worth complicating the condition.

src/backend/optimizer/plan/planner.c
Alena0704: Agree in principle, but it is not a drop-in replacement. The check would have to be recursive — select * from (select * from t) s; has no RTE_RELATION at the top level, yet it is a fully distributed query :) 
And a function RTE is not always master-only either: a function can be declared EXECUTE ON ALL SEGMENTS, and then we dispatch it to the segments.

There is also a cost side: the check runs before every query, and four field comparisons are much cheaper than walking the range table.

I'd rather keep rtable == NIL here and do the rest in a follow-up PR.

src/test/regress/expected/aggregates_optimizer.out
```diff
@@ -1073,9 +1061,6 @@ CONTEXT:  SQL function "aggf_trans" during startup
 select aggfns(distinct a,b,c)
   from (values (1,3,'foo'),(0,null,null),(2,2,'bar'),(3,1,'baz')) v(a,b,c),
        generate_series(1,3) i;
-INFO:  GPORCA failed to produce a plan, falling back to planner
```
andr-sokolov, line 1076: The query contains FROM and there is no attempt to plan it using Orca. I suggest to rewrite the PR description to make it clear which types of queries are not passed to Orca

Alena0704: Agree, I rewrote it - please check

src/backend/optimizer/plan/planner.c
```diff
@@ -5861,3 +5866,37 @@ isSimplyUpdatableQuery(Query *query)
...
+ * set-returning functions worse.  The entire optimization attempt is therefore
+ * pure overhead, and we skip it.
+ *
+ * An empty range table on its own is not conclusive, though.  A sublink can
+ * hide a subquery over a distributed relation, as in
+ * "SELECT EXISTS(SELECT 1 FROM t)", where the top level range table is still
+ * empty; such queries keep going to ORCA.  A query that is part of a utility
+ * statement (CTAS, COPY, REFRESH MATERIALIZED VIEW) is left to ORCA as well,
+ * because there it is the parent statement, not the SELECT itself, that
+ * decides what gets dispatched.
```
andr-sokolov: Why does Orca build a better plan for utility statement? Can you share an example?

Alena0704: You are right, it is not about a better plan. The point is that such a query is not local: CREATE TABLE t AS SELECT 42 still has to distribute its result row according to the target table's policy. I reworded the comment.

Alena0704:
I tried removing the parentStmtType condition and [CI ](https://github.com/open-gpdb/gpdb/actions/runs/34361763538/job/102504350676?pr=400) shows it is not safe: with optimizer=on the distribution policy of a CTAS table changes - matview gets Distributed by: (i) instead of Distributed randomly, legacy_hashops_ctas ends up hash-distributed on cdbhash_int4_ops, and create_table_distpol starts emitting using default RANDOM distribution. So I suggest not to add it in this PR.

src/backend/optimizer/plan/planner.c
```diff
@@ -5861,3 +5866,30 @@ isSimplyUpdatableQuery(Query *query)
...
+ *		Should this query be handed to the ORCA optimizer at all?
+ *
+ * A SELECT with an empty range table - "SELECT 42" or
+ * "SELECT pg_column_size('...')" - has nothing to distribute across segments,
+ * so ORCA's plan is no better than the Postgres planner's.
+ * The optimization attempt is pure overhead, and we skip it.
+ *
+ * Sublinks and utility statements are excluded.  A sublink can hide a
+ * subquery over a distributed relation, and for CTAS the result row has to be
+ * distributed according to the target table's policy.
```
andr-sokolov, line 5881: Why is it a problem that row has to be distributed according to the target table's policy?

andr-sokolov (approved):

PR #403, opened 2026-07-28
Fix the figures reported by "Rows out" in EXPLAIN ANALYZE

Alena0704:
Fix the figures reported by "Rows out" in EXPLAIN ANALYZE

Skip the segments that did not run the node instead of counting them as
workers producing zero rows, report rows per loop so that rescanned
nodes match their own row count, and take the segment numbers relative
to the slice rather than to the insts array.

Taken from GreengageDB/greengage#524, which fixed the feature while
backporting it from here.

### DIFF
--- a/src/backend/commands/explain_gp.c
+++ b/src/backend/commands/explain_gp.c
@@ -2020,59 +2020,76 @@ cdbexplain_showExecStats(struct PlanState *planstate, ExplainState *es)
 	 * Print "Rows out"
 	 */
 
-	if (gp_enable_explain_rows_out && es->analyze && ns->ninst > 0) {
-		double alltuples = 0;
-		double maxtuples = ns->insts[0].ntuples;
-		int maxseg = 0;
-		double mintuples = ns->insts[0].ntuples;
-		int minseg = 0;
+	if (gp_enable_explain_rows_out && es->analyze && ns->ninst > 0)
+	{
+		double		alltuples = 0;
+		double		maxtuples = 0;
+		int			maxseg = -1;
+		double		mintuples = 0;
+		int			minseg = -1;
+		int			workercount = 0;
 
 		for (i = 0; i < ns->ninst; i++)
 		{
 			CdbExplain_StatInst *nsi = &ns->insts[i];
+			double		rows;
 
-			alltuples += nsi->ntuples;
+			/*
+			 * The insts array is allocated with palloc0 and only the slots of
+			 * the segments that actually ran this node are filled in, so an
+			 * unset pstype means the segment did not participate.  Counting
+			 * such slots would report a bogus minimum of zero rows and drag
+			 * the average down.
+			 */
+			if (nsi->pstype == T_Invalid)
+				continue;
 
-			if (nsi->ntuples > maxtuples) {
-				maxtuples = nsi->ntuples;
+			/* Report rows per loop, like the row count on the node itself. */
+			rows = nsi->nloops > 0 ? nsi->ntuples / nsi->nloops : 0;
+
+			if (workercount == 0 || rows > maxtuples)
+			{
+				maxtuples = rows;
 				maxseg = ns->segindex0 + i;
 			}
 
-			if (nsi->ntuples < mintuples) {
-				mintuples = nsi->ntuples;
+			if (workercount == 0 || rows < mintuples)
+			{
+				mintuples = rows;
 				minseg = ns->segindex0 + i;
 			}
+
+			alltuples += rows;
+			workercount++;
 		}
 
-		double avgtuples = alltuples / ns->ninst;
-
-		if (es->format == EXPLAIN_FORMAT_TEXT)
+		if (workercount > 0)
 		{
-			/*
-			 * create a header for all stats: separate each individual stat by an
-			 * underscore, separate the grouped stats for each node by a slash
-			 */
-			appendStringInfoSpaces(es->str, es->indent * 2);
-			appendStringInfoString(es->str, "Rows out: ");
+			double		avgtuples = alltuples / workercount;
 
-			appendStringInfo(es->str,
-								 "%0.2f rows avg x %d workers, %0.f rows max (seg%d), %0.f rows min (seg%d).\n",
+			if (es->format == EXPLAIN_FORMAT_TEXT)
+			{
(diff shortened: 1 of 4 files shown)


leborchuk (approved):
LGTM, checked also original fix in GG

leborchuk (approved):
LGTM

src/backend/commands/explain_gp.c
```diff
@@ -2020,59 +2020,77 @@ cdbexplain_showExecStats(struct PlanState *planstate, ExplainState *es)
...
-		int minseg = 0;
+	if (gp_enable_explain_rows_out && es->analyze && ns->ninst > 0)
+	{
+		double		alltuples = 0;
+		double		maxtuples = 0;
+		int			maxseg = -1;
+		double		mintuples = 0;
+		int			minseg = -1;
+		int			workercount = 0;
+		double		avgtuples;
```
andr-sokolov: Let's place `avgtuples` declaration under `if (workercount > 0)`, where the variable is initialized

src/backend/commands/explain_gp.c
```diff
@@ -2020,59 +2020,77 @@ cdbexplain_showExecStats(struct PlanState *planstate, ExplainState *es)
...
-		if (es->format == EXPLAIN_FORMAT_TEXT)
+		if (workercount > 0)
 		{
-			/*
-			 * create a header for all stats: separate each individual stat by an
-			 * underscore, separate the grouped stats for each node by a slash
-			 */
-			appendStringInfoSpaces(es->str, es->indent * 2);
-			appendStringInfoString(es->str, "Rows out: ");
+			avgtuples = alltuples / workercount;
```
andr-sokolov:
```suggestion
			double avgtuples = alltuples / workercount;
```

src/backend/commands/explain_gp.c
```diff
@@ -1568,7 +1568,7 @@ static void
 cdbexplain_NodeSummary(ExplainState *es, CdbExplain_NodeSummary *ns) {
 	int			i;
 	char		aggbuf[MAXPGPATH];
-	
+
```
andr-sokolov: unrelated change

src/backend/commands/explain_gp.c
```diff
@@ -2020,59 +2020,76 @@ cdbexplain_showExecStats(struct PlanState *planstate, ExplainState *es)
...
-		if (es->format == EXPLAIN_FORMAT_TEXT)
+		if (workercount > 0)
 		{
-			/*
-			 * create a header for all stats: separate each individual stat by an
-			 * underscore, separate the grouped stats for each node by a slash
-			 */
-			appendStringInfoSpaces(es->str, es->indent * 2);
-			appendStringInfoString(es->str, "Rows out: ");
+			avgtuples = alltuples / workercount;
```
andr-sokolov: Why is the formula changed: ns->ninst is replaced with workercount?

src/backend/commands/explain_gp.c
Alena0704: Agree. Fixed

src/backend/commands/explain_gp.c
Alena0704: Added

src/backend/commands/explain_gp.c
Alena0704: Fixed

src/backend/commands/explain_gp.c
Alena0704: ns->ninst is not the number of segments that executed the node - it's the size of the insts array, segindexMax + 1 - segindexMin (explain_gp.c:1132). The array is palloc0'd and indexed by absolute segment number, so slots of segments that never ran the node stay zeroed. With ninst those empty slots were counted as workers with 0 rows, so min was 0 on a segment that never ran the node, avg was too low, and the worker count too high. 

workercount counts only the slots that were actually filled (pstype != T_Invalid), so avg/max/min and the printed worker count all refer to the same set of segments.

avg is a plain unweighted average of the per-loop row counts across the segments that ran the node.

andr-sokolov (approved):

Alena0704:
Just in case, here’s a reproduction of the problem:
```
create table ro (id int4) distributed by (id);
insert into ro select generate_series(1,10); 
set gp_enable_explain_rows_out=on;
set optimizer=off;
до фикса

postgres=# explain analyze select * from ro where id = 1;
                                                      QUERY PLAN                                                      
----------------------------------------------------------------------------------------------------------------------
 Gather Motion 1:1  (slice1; segments: 1)  (cost=0.00..3.12 rows=1 width=4) (actual time=1.284..1.285 rows=1 loops=1)
   Rows out: 1.00 rows avg x 1 workers, 1 rows max (seg0), 1 rows min (seg0).
   ->  Seq Scan on ro  (cost=0.00..3.12 rows=1 width=4) (actual time=0.015..0.016 rows=1 loops=1)
         Filter: (id = 1)
         Rows out: 1.00 rows avg x 1 workers, 1 rows max (seg0), 1 rows min (seg0).
 Planning time: 0.334 ms
   (slice0)    Executor memory: 59K bytes.
   (slice1)    Executor memory: 25K bytes (seg1).
 Memory used:  128000kB
 Optimizer: Postgres query optimizer
 Execution time: 2.804 ms
(11 rows)

после фикса
postgres=# explain analyze select * from ro where id = 1;
                                                       QUERY PLAN                                                       
------------------------------------------------------------------------------------------------------------------------
 Gather Motion 1:1  (slice1; segments: 1)  (cost=0.00..3.12 rows=1 width=4) (actual time=28.323..28.324 rows=1 loops=1)
   Rows out: 1.00 rows avg x 1 workers, 1 rows max (seg-1), 1 rows min (seg-1).
   ->  Seq Scan on ro  (cost=0.00..3.12 rows=1 width=4) (actual time=0.042..0.043 rows=1 loops=1)
         Filter: (id = 1)
         Rows out: 1.00 rows avg x 1 workers, 1 rows max (seg1), 1 rows min (seg1).
 Planning time: 0.569 ms
   (slice0)    Executor memory: 59K bytes.
   (slice1)    Executor memory: 25K bytes (seg1).
 Memory used:  128000kB
 Optimizer: Postgres query optimizer
 Execution time: 31.154 ms
(11 rows)
```

PR #404, opened 2026-07-28
Fix distribution by resjunk column

Alena0704:
Fix distribution by resjunk column

In the PostgreSQL planner, some queries require auxiliary resjunk columns in
the targetlist.

These columns are necessary for query execution but do not appear in the
resulting relation.

Our logic for CTAS expressions, however, did not account this case.
If a query didn't contain an explicit DISTRIBUTED BY clause,
and other columns in the targetlist were not hashable,
the planner could use resjunk column in the distribution policy,
resulting in a segmentation fault.

To fix this issue, simply filter out resjunk columns.

Also, make attribute selection more efficient, as the targetlist
was re-examined for each entry.

Cherry-pick from GreengageDB/greengage#318

### DIFF
--- a/src/backend/cdb/cdbmutate.c
+++ b/src/backend/cdb/cdbmutate.c
@@ -465,39 +465,38 @@ apply_motion(PlannerInfo *root, Plan *plan, Query *query)
 					 */
 					if (!targetPolicy)
 					{
-						int			i;
 						List	   *policykeys = NIL;
 						List	   *policyopclasses = NIL;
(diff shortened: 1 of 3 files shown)


andr-sokolov:
the base patch has been got from GG 7.x. Can we get the patch from 6.x?

Alena0704:
> the base patch has been got from GG 7.x. Can we get the patch from 6.x?

Done

andr-sokolov (approved):

PR #411, opened 2026-08-18
CVE backports

reshke:
- [ ] Document changes


leborchuk:
LGTM, the overall backport is good enough. 

But let's discuss some details, I cannot decide if they are worth fixing or not:

1. https://www.postgresql.org/support/security/CVE-2026-6470/ was partially fixed. 

it does not include 1a358b8f2a2

```
commit 1a358b8f2a28aaf3a7bfc8cf710de7a39c98436d
Author: Nathan Bossart <nathan@postgresql.org>
Date:   Mon Aug 10 06:38:36 2026 -0700

    Check for USAGE privilege on types used by stored expressions.

    This omission allowed roles without USAGE on a type to create
    stored expressions that depend on it, which could prevent the owner
    from changing the type later.

    The checks deliberately live in the command paths rather than the
    dependency-recording routines.  Those routines also run whenever
    the server re-derives an existing expression, and re-checking there
    would break routine maintenance for an owner who has since lost
    USAGE on a type its objects already reference.  (Checking in the
    dependency-recording routines would also require additional
    CommandCounterIncrement() calls to avoid spurious errors.)

    The addition of a parameter to AlterDomainAddConstraint() breaks
    ABI compatibility, but we are unaware of any impacted third-party
    code.

    Reported-by: Noah Misch <noah@leadboat.com>
    Author: Nathan Bossart <nathandbossart@gmail.com>
    Reviewed-by: Noah Misch <noah@leadboat.com>
    Reviewed-by: Tom Lane <tgl@sss.pgh.pa.us>
    Reviewed-by: Robert Haas <robertmhaas@gmail.com>
    Security: CVE-2026-6470
    Backpatch-through: 14
```

Maybe we should take this PR too?

2. Something strange with pgcrypto tests.

`pgp-encrypt_1.out` is upstream file while it should be copied from https://github.com/open-gpdb/gpdb/blob/OPENGPDB_STABLE/contrib/pgcrypto/expected/pgp-encrypt.out 

Now `pgp-encrypt_1.out` contains tests never existed in GPDB - for example `s2k-count`. 

But the CI is green. It means something wrong with tests and CI both.

The same with files pgp-decrypt / pgp-pubkey-decrypt. Upstream also touched pgp-decrypt_1.out and pgp-pubkey-decrypt_1.out; this tree has no _1 files at all.

I do not know if we should create pgp-decrypt_1.out since the issue could be with support Blowfish by openSSL. But it's worth investigating.

3. trgm_gist.c - no need to changed idents. Both else branches were de-indented by one level while making the one-token fix:

```
-                     else
-                             size_alpha = SIGLENBIT - sizebitvec(
+             else
+                     size_alpha = SIGLENBIT - sizebitvec(
```

4. Levenshtein fix ships with zero tests. Upstream added 5 test cases in contrib/fuzzystrmatch/{sql,expected}/fuzzystrmatch.*; this tree's fuzzystrmatch has no regress suite at all, so they were dropped. Why? Maybe it should worth add it.

5. amcheck fixes should be checked once again, I think we missed some sections while porting it. 

Upstream PG 14's contrib/amcheck/sql/check_btree.sql already contained a varlena_bug test block before the CVE fix landed. It was added in 2024 by two unrelated commits that fixed a real amcheck bug:

- 5df5d9cd7ea — amcheck: Support for different header sizes of short varlena datum (2024-03-23)
- 4e8529da48b — amcheck: Normalize index tuples containing uncompressed varlena (2024-03-23)
- (and later af09b18cbad, 2026-06-14, fixing that same code)

Those commits introduced bt_normalize_tuple() in verify_nbtree.c, and the varlena_bug test exists to exercise it.

The CVE-2018-1058 commit — 39d792040c7 — sits directly above that block.

There is no `bt_normalize_tuple` in current code. Most probably we should also take 5df5d9cd7ea fix


Also, August 2026 batch contains other fixes - 

CVE-2026-14662 tsvector/tsquery overflow (2 commits) · 14666 plan-cache invalidation after role changes · 14668 scalarineqsel() const datatype · 14670 PL/Perl tied arrays · 14677 palloc_array() in pltcl/plperl · 14679 FUNC_MAX_ARGS arrays (2 commits) · 14680 type internal functions (2 commits) · 15741 EXTRACT() deparse injection · 16239 portal type cross-check · 16241 ecpg bytea OOB write · 18024 ascii() OOB read · 6464 psql \if/COPY FROM STDIN (2 commits)

Maybe we also could take some of them?

PR #418, opened 2026-08-27
Add TEA

andr-sokolov:
Add the --enable-tea option to configure. If this option is not specified, TEA
is not built.
Add the gpcontrib/tea directory with submodules for TEA and Arrow. Arrow is
needed as a dependency of TEA. The OPENGPDB_GP6 branch of TEA is used, because
it contains patches to build TEA without gRPC and does not requier installed GP.
When make is executed in gpcontrib/tea, it downloads boost-1.81.0.tar.gz.
The path to boost-1.81.0.tar.gz is passed to cmake in the ARROW_DEPS variable.
TEA requiers cmake >= 3.25, but Ubuntu 22.04 has 3.22.1. So we should download
a newer cmake version.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -5,3 +5,11 @@
 	path = gpcontrib/yezzey
 	url = https://github.com/yezzey-gp/yezzey.git
 	branch = v1.8
+[submodule "gpcontrib/tea/tea"]
+	path = gpcontrib/tea/tea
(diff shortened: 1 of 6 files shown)


leborchuk (approved):
LGTM, disabled by default, do not affect current CI/CD workflow

PR #424, opened 2026-09-12
gp_stats_collector: opt-in EXPLAIN (FORMAT JSON) plans, extension 1.2

Sanikadze:
#### Problem

Consumers that draw the plan as a tree (yagpcc's Grafana panel via PEV2) need it as data. The text plan does not parse reliably: structure is in indentation, Greenplum adds its own annotations (`Gather Motion 2:1 (slice1; segments: 2)`, `Rows out: ...`, ORCA cost lines), and PEV2 fromText drops Motion nodes. EXPLAIN (FORMAT JSON) comes from the same tree walk and is stable across minor versions.

Text plans are unchanged and always collected. plan_id is still the hash of the normalized text plan.

#### Extension 1.1 -> 1.2

gpsc.__log is built positionally from log_tbl_desc, so the layout changes.

- `gp_stats_collector--1.1--1.2.sql`: `ALTER TABLE ... ADD COLUMN` x2, gpsc.log view recreated.
- Fresh 1.2 and upgraded 1.1 tables have the same layout. utility is looked up by name, not assumed last.
- A 1.2 .so on a 1.1 table does not write misaligned rows: insert_log checks the attribute count, warns once per backend with a HINT to run `ALTER EXTENSION gp_stats_collector UPDATE`, and skips table logging until then. UDS logging is unaffected.
- Install .so, extension/*.sql and .control on all hosts (ALTER EXTENSION is dispatched to segments). New .so needs a restart.

#### Compatibility

- Default off: existing installations get only the GUC and the 1.2 files.
- Older agent keeps 15/16 as unknown protobuf fields and ignores them.
- JSON payload is 3-5x the text plan, hence opt-in.


Sanikadze:
Confirmed, and it is worse than the review says: plain `EXPLAIN (ANALYZE, VERBOSE, FORMAT JSON)` in psql is already invalid JSON with `explain_memory_verbosity` other than `suppress` (`cdbexplain_showExecStats` appends the `Memory:` line without checking `es->format`, explain_gp.c:1865) and with `gp_enable_explain_allstat = on` (`Allstat` is opened as an object and each segment as an unlabeled array, explain_gp.c:2112). `gp_enable_explain_rows_out`, `gp_enable_explain_node_summary` and both optimizers are fine.

Pushed 4 commits on top:

- both GUCs are saved, forced off while `analyze_json` is rendered and restored right after (643a1854). `analyze_text` is not affected.
- the `gpsc.__log` layout check now runs before the payload is converted (b54d5b32).
- `gpsc_json_plan` regress: a distributed join with a Motion node and nested `Plans` (8b5388fa), `explain_memory_verbosity = detail` (fails on the old code with `Token "Memory" is invalid`), and all four EXPLAIN GUCs at once with both payloads parsed through `::json` (eed0ba67).

The core EXPLAIN bug is separate from this PR and worth its own fix in explain_gp.c. Related yagpcc side: open-gpdb/yagpcc#56.

PR #428, opened 2026-09-18
Add DML operation logging to access_log

MutableFire:
Extend `access_log` to log write access for `INSERT`, `UPDATE`, and `DELETE`
operations.

A new `operation` field distinguishes `read` events produced by Seq Scan
initialization from `write` events produced for DML target tables.

The existing Seq Scan logging behavior is preserved. Regression tests cover
heap, AO, AOCO, partitioned tables, direct dispatch, and user/session metadata.


gpcontrib/access_log/access_log.c
```diff
@@ -40,60 +42,198 @@ write_to_log(const char* str)
...
+
+	/*
+	 * 19 ("YYYY-MM-DD HH:MM:SS")
+	 * + 1 (".")
+	 * + 6 (microseconds)
+	 * + 1 (" ")
+	 * + 5 (time zone abbreviation)
+	 * + 1 ("\0").
+	 */
+	char			buf[19 + 1 + 6 + 1 + 5 + 1];
```
Vlasdislav, line 58: The transcript is in the comments, maybe you should write down the total number?

MutableFire: Done.

Alena0704:
Why test output changed (in gpcontrib/access_log/expected/access_log.out)? maybe you should add about it in the commit message?

gpcontrib/access_log/access_log.c
andr-sokolov: It will be not obvious to check that the total number is correct

gpcontrib/access_log/access_log.c
```diff
@@ -59,41 +68,174 @@ access_log_init_scan_hook(Relation currentRelation)
...
+{
+	StringInfoData	buffer;
+
+	if (!is_loggable_relation(currentRelation))
+		return;
+
+	initStringInfo(&buffer);
+
+	append_log_time(&buffer);
+	appendStringInfoChar(&buffer, ',');
```
Alena0704, line 122: Look at line 63: the pg_strftime format "%Y-%m-%d %H:%M:%S        %Z," already ends with a comma. This line adds another one, so every record gets an empty field after the timestamp. The line then has 6 CSV fields while the access_log external table has only 5 columns, and it can't be read correctly.

gpcontrib/access_log/access_log.c
```diff
@@ -59,41 +68,174 @@ access_log_init_scan_hook(Relation currentRelation)
...
+
+static void
+access_log_ExecutorStart(QueryDesc *queryDesc, int eflags)
+{
+	if (next_ExecutorStart_hook)
+		next_ExecutorStart_hook(queryDesc, eflags);
+	else
+		standard_ExecutorStart(queryDesc, eflags);
+
+	access_log_dml(queryDesc);
```
Alena0704: should we check eflags & EXEC_FLAG_EXPLAIN_ONLY?

gpcontrib/access_log/access_log.c
MutableFire: Done.

gpcontrib/access_log/README.md
```diff
@@ -1,11 +1,27 @@
...
-the extension you can use the shared_preload_libraries GUC or the LOAD command.
+The extension logs access to Greenplum tables and partitions.
 
-If you want to register Seq Scans on segments only you should load access_log
-on segments and don't load on master:
+For read operations, it logs when, which user and in which session initializes
+a Seq Scan on a table or partition.
+
+For write operations, it logs when an `INSERT`, `UPDATE`, or `DELETE` operation
+starts on a target table. For a partitioned target, the root target table is
```
Alena0704: This is only true for INSERT. For UPDATE/DELETE on a partitioned table, I think the partitions will be logged instead of the root table. Could you add a test for UPDATE/DELETE on t_heap_part and check it?

gpcontrib/access_log/access_log.c
Alena0704: The calculation doesn't count the comma at the end of the format on line 63, so the buffer is 1 byte short for a 5-char time zone abbreviation (e.g. Asia/Kathmandu). Then pg_strftime doesn't write \0 and we read past the buffer, so garbage gets into the log.

The bug reproduction:
```
gpconfig -c log_timezone -v Asia/Kathmandu
gpstop -u
```
```sql
create table t_repro (a int) distributed by (a);
insert into t_repro select generate_series(1, 100);
load 'access_log';
select count(*) from t_repro;

-- read the log through the access_log external table from the test
select * from access_log;
```
```
ERROR:  invalid byte sequence for encoding "UTF8": 0xe2 0xef 0xe1  (seg0 slice1 127.0.1.1:9802 pid=3932479)
```

gpcontrib/access_log/access_log.c
```diff
@@ -59,41 +68,174 @@ access_log_init_scan_hook(Relation currentRelation)
...
+		   currentRelation->rd_rel != NULL &&
+		   currentRelation->rd_rel->relkind == RELKIND_RELATION;
+}
+
+static void
+log_relation_access(Relation currentRelation, const char *operation)
+{
+	StringInfoData	buffer;
+
+	if (!is_loggable_relation(currentRelation))
```
Alena0704: This check is also applied to `read` now (`access_log_init_scan_hook` calls `log_relation_access`), so Seq Scan on relations with other relkinds, e.g. materialized views, is not logged anymore. Previously every Seq Scan was logged. Is it intended? If not, maybe move the check to `access_log_dml` only.

The reproduction:
```sql
create table t_repro (a int) distributed by (a);
insert into t_repro select generate_series(1, 100);
create materialized view mv_repro as select * from t_repro distributed by (a);

load 'access_log';
explain (costs off) select count(*) from mv_repro;
select count(*) from mv_repro;
```
```
                   QUERY PLAN
------------------------------------------------
 Aggregate
   ->  Gather Motion 3:1  (slice1; segments: 3)
         ->  Aggregate
               ->  Seq Scan on mv_repro
```
Before this PR, `pg_log/access.log` on the segments:
```
2026-09-21 22:13:03.060496 UTC,alena,con75,public.mv_repro
2026-09-21 22:13:03.060549 UTC,alena,con75,public.mv_repro
2026-09-21 22:13:03.060556 UTC,alena,con75,public.mv_repro
```
With this PR nothing is logged, `pg_log/access.log` is not even created.

Alena0704 (changes requested):

gpcontrib/access_log/access_log.c
```diff
@@ -40,60 +42,202 @@ write_to_log(const char* str)
...
+		access_log_dml(queryDesc);
+}
+
+void
+_PG_init(void)
+{
 	next_init_scan_hook = init_scan_hook;
 	init_scan_hook = access_log_init_scan_hook;
 
-	inited = true;
```
andr-sokolov, line 98: This change is unrelated to adding DML logging

MutableFire: Agreed. We discussed it earlier and decided that this initialization guard was unnecessary, so it was removed to keep the code cleaner. I can mention in the PR description that minor unnecessary checks were also removed as part of the cleanup.

andr-sokolov: Let's do it in another PR to don't mix unlinked changes

Alena0704 (approved):

gpcontrib/access_log/access_log.c
MutableFire: Done.

gpcontrib/access_log/access_log.c
```diff
@@ -40,49 +42,196 @@ write_to_log(const char* str)
...
+static void
+log_relation_access(Relation currentRelation, const char *operation)
+{
+	StringInfoData	buffer;
+
+	if (currentRelation == NULL ||
+		currentRelation->rd_rel == NULL)
+		return;
+
+	initStringInfo(&buffer);
```
andr-sokolov, line 119: Can we use stack memory instead of heap?

gpcontrib/access_log/access_log.c
```diff
@@ -40,49 +42,196 @@ write_to_log(const char* str)
...
+{
+	EState		   *estate;
+	ResultRelInfo *resultRelInfo;
+	Oid			   *loggedOids;
+	int				numResultRelations;
+	int				numLoggedOids = 0;
+	int				i;
+
+	if (queryDesc == NULL)
+		return;
```
andr-sokolov: Why do we need this check?

gpcontrib/access_log/access_log.c
```diff
@@ -40,49 +42,196 @@ write_to_log(const char* str)
...
+	resultRelInfo = estate->es_result_relations;
+	numResultRelations = estate->es_num_result_relations;
+
+	if (resultRelInfo == NULL ||
+		numResultRelations <= 0)
+		return;
+
+	loggedOids =
+		(Oid *) palloc(sizeof(Oid) * numResultRelations);
+
```
andr-sokolov, line 193: In which case numResultRelations is more that 1 and we need to allocate memory for array?

gpcontrib/access_log/access_log.c
```diff
@@ -40,49 +42,196 @@ write_to_log(const char* str)
...
+		Relation	currentRelation =
+			resultRelInfo[i].ri_RelationDesc;
+		Oid			oid;
+
+		if (!is_loggable_relation(currentRelation))
+			continue;
+
+		oid = RelationGetRelid(currentRelation);
+
+		if (oid_seen(loggedOids, numLoggedOids, oid))
```
andr-sokolov, line 205: In which case relation oid is duplicated?

gpcontrib/access_log/expected/access_log.out
```diff
@@ -439,9 +453,9 @@ select * from t_aoco;
 select * from show_log(:'before_query');
  gp_segment_id | date_ok | user_ok | sess_ok |      tbl      
 ---------------+---------+---------+---------+---------------
+             0 | t       | t       | t       | public.t_aoco
              1 | t       | t       | t       | public.t_aoco
              2 | t       | t       | t       | public.t_aoco
-             0 | t       | t       | t       | public.t_aoco
```
andr-sokolov: There is no need to change the tuples order

MutableFire: Done.

gpcontrib/access_log/access_log.c
MutableFire: No, using stack memory is not recommended here. The length of the log record is variable (depends on table names, user names, etc.), and coding standards prefer StringInfo/palloc for dynamic data to safely handle variable lengths and integrate with the memory context system.

gpcontrib/access_log/access_log.c
MutableFire: No need. Deleted.

gpcontrib/access_log/access_log.c
MutableFire: numResultRelations is greater than 1 when a single DML statement targets multiple physical relations simultaneously. This typically happens with INSERT, UPDATE, or DELETE on partitioned tables where the executor initializes separate ResultRelInfo structures for different leaf partitions.

gpcontrib/access_log/access_log.c
MutableFire: Duplicates can occur during complex DML operations on partitioned tables where the executor might reference the same target relation multiple times within es_result_relations. The check prevents logging the same table twice for a single command.

andr-sokolov: Is there an example of such query in the added tests?

gpcontrib/access_log/access_log.c
andr-sokolov: table name length is limited to 64 bytes. stack memory allocation is much more faster that heap

MutableFire: I decided to keep StringInfo because it is more readable and reduces the risk of buffer overflow errors if log formats change in the future. The performance difference between stack and heap allocation is negligible for administrative logging, which is not a performance-critical path.

gpcontrib/access_log/access_log.c
MutableFire: It may occur in complex DML scenarios on partitioned tables, a simple regression test does not explicitly trigger this case, the check ensures robustness against future kernel changes.

gpcontrib/access_log/access_log.c
```diff
@@ -75,14 +77,58 @@ access_log_init_scan_hook(Relation currentRelation)
...
 
 	if (next_init_scan_hook)
 		next_init_scan_hook(currentRelation);
 }
 
+static void
+access_log_ExecutorStart(QueryDesc *queryDesc, int eflags)
+{
+	EState		   *estate;
+	ResultRelInfo *resultRelInfo;
```
andr-sokolov: align please
```suggestion
	ResultRelInfo  *resultRelInfo;
```

